PR #9, opened 2015-01-10
Changed keys for showing and clearing PQ renderers to [ and ]

Gerry1135:
Various people have been complaining that it is too easy to accidentally hit Ctrl-S, especially during EVA or landing, so I have changed the keys from Ctrl-S and Ctrl-C to Ctrl-[ and Ctrl-]

### DIFF
--- a/Kopernicus/RuntimeUtility.cs
+++ b/Kopernicus/RuntimeUtility.cs
@@ -151,7 +151,7 @@ namespace Kopernicus
 			}
 
 			// If we want to debug the locations of PQ nodes
-			if (Input.GetKeyDown (KeyCode.LeftBracket) && Input.GetKey (KeyCode.LeftControl)) 
+			if (Input.GetKeyDown (KeyCode.Semicolon) && Input.GetKey (KeyCode.LeftControl)) 
 			{
 				// New list for the debugger 
 				List<GameObject> renderers = new List<GameObject>();
@@ -181,7 +181,7 @@ namespace Kopernicus
 			}
 
 			// If we want to clean up 
-			if (Input.GetKeyDown (KeyCode.RightBracket) && Input.GetKey (KeyCode.LeftControl)) 
+			if (Input.GetKeyDown (KeyCode.Slash) && Input.GetKey (KeyCode.LeftControl)) 
 			{
 				// Disable all of the renderers in the list
 				foreach(GameObject r in quadSurfaceNormalRenderers)
--- a/Kopernicus/RuntimeUtility.cs
+++ b/Kopernicus/RuntimeUtility.cs
@@ -119,12 +119,14 @@ namespace Kopernicus
 		}
 		
 		/**
-		 * If Control-P are pressed, dump the PQS data of the current live body
+		 * If Mod-P are pressed, dump the PQS data of the current live body
 		 **/
 		public void Update()
 		{
-			// Print out the PQS state
-			if( Input.GetKeyDown( KeyCode.P ) && Input.GetKey( KeyCode.LeftControl ) )
+            bool isModDown = GameSettings.MODIFIER_KEY.GetKey();
+            
+            // Print out the PQS state
+            if (Input.GetKeyDown(KeyCode.P) && isModDown)
 			{
 				// Log the state of the PQS
 				Utility.DumpObjectFields(FlightGlobals.currentMainBody.pqsController, " Live PQS ");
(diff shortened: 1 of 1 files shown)


Gerry1135:
Changed the keys to Ctrl-; (semicolon) and Ctrl-/ (slash) to avoid vessel switching.

MonArchangel:
HI,

 

Ctl-[ is not such a good idea as this is equivalent to pressing the ESC key.... 

 

To confirm this, refer to the included ACII character table where CTL-@ is equivalent to the NUL character 

 

0x5B = [       -> 0101  1011

0x40 = @     -> 0010 1000

0x1B = ESC -> 0001 1011

 

0x5B – 0x40 = 0x1b = 27 decimal .

 

In effect, when the IRQ 9 interrupt is intercepted, if the CTL key is pressed, any other key value will have 64 decimal subtracted from its value.

 



 

Thus selection of this key combination might prove problematic for users particulary as the ESC key is intended to pop up the Game Paused dialogue.

 

Good luck,

 

Bruce Knowles

BryceSchroeder:
Any suggestions for alternative keys to [ and ]?

MonArchangel:
Oops... this mail arrived just seconds after I shipped my earlier one...

 

Please ignore it.

MonArchangel:
Hi Bryce,

 

I have mulled over your question a while, spending a bit of time considering some choices that might prove less of a problem. It occurred to me to use keys whose values that were less than 64 decimal.  The reason here is that because the 64 value that gets subtracted, I reason that it is probable that a signed value will be generated. In its unsigned form, this well mean an ASCII value in the upper half of the character  set lying in the space where all the graphics characters are. 

 



 

On the presumption that 2’s complement arithmetic is implemented at the hardware layer.

 

If we select the keys ‘-‘ and ‘=’ for the purposes of this discussion, 

‘-‘ = 45 decimal; Subtracting 64 = -19 which in binary will be 1110 1101 or 0xED.  

 

To convert this value to a decimal value in the Extended ASCII code range all we need to do is add 256 to derive the resultant character.

 

So  CTL->‘-‘ = -19; adding 256 yields 237 which I think is the character called Rho.

Similarly; CTL->’=’ =  -3; adding 256  yields 253 which I think is the character called Zeta.

 

The generated key values are thus in the general case far removed from what is currently in use within the game and therefore very unlikely to conflict with other key values with alternate uses.

 

This thought process may be of some use to you in choosing a more proper choice of keys supposing that you wish to change from the current  CTL-‘;’ and CTL-‘/’ that you have swapped to.

 

Good luck,

 

Bruce Knowles

Gerry1135:
I'm not sure about Linux and MacOS but this definitely isn't how the keyboard handling works in Windows Unity.  If you hit Ctrl-[ then Input.GetKeyDown(KeyCode.Escape) does not return true.  The fact that the [ and ] keys switch vessel is a problem though as the KSP code doesn't check that Ctrl is not pressed when testing the bracket keys (or any other keys for that matter).  Semicolon and Slash are not used by the default stock config so these don't cause any undesirable side-effects though the left Ctrl key is used to decrease throttle which could be undesirable.  I think the code should actually be changed to use GameSettings.MODIFIER_KEY.GetKey() instead of Input.GetKey(KeyCode.LeftControl) which would change the keys to use Alt on Windows and RCtrl on Linux/Max instead of LCtrl.  This wouldn't mess with the vessel's throttle.

MonArchangel:
AHA,

 

It seems that I was a bit premature in my initial mail to you. I have forgotten that you were building code for a .NET based product where a number of more modern techniques and software mechanisms now exist.

 

I can assure you that the discussion I raised is for the most part OS independent and revolves more around the older hardware mechanisms that were implemented on computer equipment including PC’s running all known versions of MS-DOS, DR-DOS and Windows. Of course, you need to be working in languages with operate a lot closer to the hardware layer to get exposed to these ideas I suppose. Some experience in C or C++ (or even Intel/AMD/Motorola chipset Assembler tools) will demonstrate this.

 

In contrast though, as a consequence of keyboards (for example) now being virtualised coupled with the complications of international character sets, key layouts on keyboards (virtual or otherwise) new approaches were necessary to make it easier for developers to contend with. In the case of the function you named, this approach is an implementation of a mapping of the name of a key that you are interested in to its physical location on a ‘virtual’ keyboard. It is not the ASCII value of that key nor for that matter is it the equivalent Locale specific  Unicode value.

 

In short though, your approach is a much less restrictive  method than mine is. 

 

You are also correct in asserting that given the tools that you need to use in Unity, the risk I nominated cannot occur. Please accept my apologies for this misdirection.

 

Best wishes,

 

Bruce Knowles

PR #12, opened 2015-03-01
Added ability to generate the "Kopernicus"-planet + the config-loaded PSystem

ThomasKerman:
Just moved some code around, added a check for "Kopernicus = true" in System.cfg, and made the path to the textures a little bit nicer. Oh, and updated ModuleManager. :smiley:


ThomasKerman:
Removed body-rearranging in RD to fix Star-Bug. Kopernicus was VERY buggy, whey you was trying to create a star. :D

Olympic1:
Will it then be possible to create stars again?

ThomasKerman:
Yes, star creation is working now. ;)

ThomasKerman:
Yes, star creation is working now. ;)

Gravitasi:
Hello.

I've checked the changes, and... I hate to say this... but personally, I can't say that I like what you did here.

There are some things that I would like to address:
1.  I believe "Kopernicus" planet was meant only to be used by developers, so they could try things out. It never meant to be released along with official, (pre-)release version. So, adding Kopernicus planet along with planets in config may be out-of-scope with the plugin's original intention. I could be wrong on this matter, though. By the way, of course, no one can stop you making your own version that does it. 
2.  The weird-named "PluginData" folder was there for a reason: so the textures inside wouldn't get automatically loaded by KSP GameDatabase.
3.  It's nice that we can get stars working again. But, I don't agree that the way we make it work is by removing some features. Nevertheless, you seem successfully pinpoint where the problem is. I haven't checked it out myself, but considering the fact that you removed a recursive function... stack overflow perhaps?

Sorry if I seem rude or something, I never intended to hurt anyone's feeling. This is just my personal opinion; the final decision still goes to the Kopernicus team.

Thank you for your attention.

ThomasKerman:
No problem Gravitasi, thanks. ;)
I see, if I can fix the function, that rearrages the bodys, and reintroduce the PluginData folder. :)

Teknoman117:
I agree with Gravatsi, which is why I didn't merge this.  The Kopernicus planet actually will end up as Mars in a future version of Real Solar System, as that's where the content came from.  I see that you emailed me Thomas, I'll hopefully get back to you later today.

ThomasKerman:
Okay, thanks! I think this can be closed. :)

PR #13, opened 2015-03-19
StarFix v2

ThomasKerman:
It's working now. :D The problem was, that Stars actually don't have CelestialBody, so I switched to the generated OrbitDriver. :)


Teknoman117:
I'll check this when I get home today.  Looks solid from a quick glance though.

BryceSchroeder:
Thanks.

PR #33, opened 2015-04-01
Added some PQS mods

Gravitasi:
I have added some loader for PQS mods. Not really sure about what to put in ModifierRequirements, since there are no stock planets that use any of the mods. Nevertheless, so far they work correctly.

I will add some more in the future.

Thanks.


Teknoman117:
I actually just discovered that setting the requirements isn't necessary, for the PQS mods set them themselves when they are started.

ThomasKerman:
I've never specified them :D
Also, could you look into my development branch Tekno? I've added "some" Mods. :)

Teknoman117:
Already did

Teknoman117:
Currently working on fixing the SOI bug.  Also, at some point we are going
to have to go through and document everything.  And provide examples (with
pictures)....

PR #40, opened 2015-05-23
Kopernicus 0.0.7

ThomasKerman:
Yeah, I think that we just need to update the examples, before we can publish the 0.0.7 version of Kopernicus. ^^ If you have questions, you can find us (NathanKell and me) in the Koepernicus - channel. :)

Thomas


NathanKell:
@ThomasKerman um @Felger just found an issue with changing scenes. So I wouldn't quite go ahead yet.

NathanKell:
Confirmed NRE issue, working on it. Closing until fixed.

NathanKell:
As far as we know, things are fixed! :)

NathanKell:
@Teknoman117  @BryceSchroeder  yep, after continuing testing things look solid. The one thing I'm slightly uncomfortable with is force-running GC after unloading a map, but since Kopernicus loads directly rather than via a coroutine that is necessary to do any GC between body loadings.

PR #43, opened 2015-06-06
Allow PQS cache file location to be explicitly specified

dbent:
This allows `Body` configuration nodes to optionally specify a `cacheFile` property, which is a `GameData` relative path to the file to be used for the PQS cache. If not specified, the current behavior is used.

This will allow mods to distribute/build the cache files inside their own directories, making mod management easier for users.

I tested this by downloading the latest Kopernicus release, replacing the assembly with one built from this version of the code, editing the Kopernicus example planet to have a `cacheFile` property, and then observing Kopernicus writing the cache file to the specified location.

### DIFF
--- a/Kopernicus/Configuration/Body.cs
+++ b/Kopernicus/Configuration/Body.cs
@@ -65,6 +65,9 @@ namespace Kopernicus
             }
             private string _name;
 
+            [ParserTarget("cacheFile", optional = true)]
+            public string cacheFile { get; set; }
+
 			[ParserTarget("cbNameLater", optional = true)]
             private string cbNameLater
             {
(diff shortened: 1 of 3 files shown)


dbent:
As a concrete example of how this would be helpful see KSP-CKAN/NetKAN#1533. The Sigma mod modifies one of OPM's planets and provides a new cache file, which it does right now by overwriting it. This means in order to properly support it in CKAN, the planet has to split off from OPM and two mods have to be made to provide it. This is worse for players that perform a manual installation, because if they wish to uninstall Sigma they have to remember to remove the cache file it provides and replace it with the one from OPM.

With this change, Sigma would just have to provide a MM patch editing the cache file location, no overwrite necessary. On uninstallation the patch file would disappear and Kopernicus would revert to using the default cache file or whatever OPM points to.

ThomasKerman:
Thanks!

dbent:
Thanks. FYI, you can set GitHub's [default branch](https://help.github.com/articles/setting-the-default-branch/), which would be used for things like pull requests and what branch people are on when they are on after a clone.

ThomasKerman:
Yes, but I have no access to that section, so I can't change it. :)

dbent:
Ah, carry on. :)

PR #53, opened 2015-07-05
CelestialBodyTransform.planetFade configuration

HappyFaceIndustries:
I didn't see anywhere in the code that allowed configuration of the PQS's fade values, and since they were locked at 100km and 110km in the PQSLoader constructor, whenever the scaledspace fader values were set to anything higher than 110km, the PQS would fade away, and then there would be no planet until the scaled version faded in.

Also, for some reason, my previous pull request was bundled up with this one, but it shouldn't adversely affect anything.


ThomasKerman:
Thanks! 
P.S. Next time, please commit to development, master is only for releases. :]

HappyFaceIndustries:
Thanks, I didn't know.

BryceSchroeder:
Sorry Thomas, didn't realize that applied to bugfixes too, which is why I
accepted it.

PR #61, opened 2015-07-12
Added land scatter loader and other stuffs

Gravitasi:
Changes/Additions:
- Land scatter loader, both for custom and stock scatter
- RGB color parser
- Added ring mesh step parameter
- Added some atmospheric parameters
- Custom PQS mod fix

In case you're curious, here's a screenshot of the custom land scatter that I've been experimenting on Hodor: http://i.imgur.com/YhCOoFY.jpg

Let me know whether I should update Hodor example with the custom land scatter config that I've made. I'll prepare it right away.

Cheers!


ThomasKerman:
Woo, Gravitasi is back! \o/
Would be cool if you would update the example. :)

Gravitasi:
Haha, okay then, give me a few minutes... :)

Gravitasi:
Done! :D

ThomasKerman:
Thanks! I will test this and leave the pull open. :)

Teknoman117:
Any particular reason for the meshStruct in the obj loader over outputting a unity mesh?

Gravitasi:
@Teknoman117: The meshStruct is only used internally in the static class (ObjImporter), as far as I can tell. Mesh loading is done by using ImportFile function, which returns a Unity mesh.

I'm not sure why the original author create that internal struct, but I decided not to change the original code too much.

PR #73, opened 2015-08-02
Add a MuParser to BuiltinTypeParsers

HappyFaceIndustries:
Just adds a simple .mu file parser that gets a mu file from the GameDatabase.


ThomasKerman:
Seems to be useful, thanks!

Teknoman117:
We should probably roll this into a generic mesh type parser.  We have an obj mesh loader already, so it would seem to make sense to make a generic mesh type parser.

PR #86, opened 2015-09-04
Use double vs int

Olympic1:
You are dividing two int's wich each other that become a new int, in this case it is 0. If you'ld make one or both of the int's a double, it will use the power of 0.33 instead of just 0.

I had this problem in my own code too.

Fixes my comment on e70b466964c6bd20bca962a72d31cf990d0f32c5


ThomasKerman:
Merged

PR #88, opened 2015-09-07
Check if ocean is null in KopernicusBuoyancy & VS2013 build fixes

### LINKED ISSUE #87: OceanUtility.cs gives Frame drops
The discussion about the frame drops: 
* http://forum.kerbalspaceprogram.com/threads/114649?p=2169273#post2169273
* http://forum.kerbalspaceprogram.com/threads/114649?p=2177577#post2177577

I downloaded the source and opened it in VS, trying to see where the problems are starting. I narrowed it down to the `KopernicusBuoyancy` class.

Trying to compile the project also gave me some errors on [Line 46](https://github.com/Kopernicus/Kopernicus/blob/development/Kopernicus/RuntimeUtility/OceanUtility.cs#L46) of the `OceanUtility.cs`. Going to try to fix this.

Josh6680:
So I had a bit of extra time today and figured I'd try and help out a bit.

First of all I fixed a few build errors in VS2013 (since that's what I use).
See commit c9bbb04 for details on that.

Secondly I added a check to see if the ocean is null in the KopernicusBuoyancy FixedUpdate function before it tries to actually do anything with the ocean.

This should fix some [NullReferenceExceptions that were reported](http://forum.kerbalspaceprogram.com/threads/114649-1-0-4-Kopernicus-Beta-%280-3-3%29-August-28?p=2177577&viewfull=1#post2177577) in the forum thread.

This should also fix #87 as the frame drop was most likely coming from the NullReferenceException error spam.

**Note:** I haven't tested this change, if someone could test this change and confirm it works, and that it fixes the NullReferenceExceptions, that would be great, but I'm almost certain it will ;)

### DIFF
--- a/Kopernicus/Configuration/Parser/BuiltinTypeParsers.cs
+++ b/Kopernicus/Configuration/Parser/BuiltinTypeParsers.cs
@@ -656,7 +656,7 @@ namespace Kopernicus
                     if (mesh == null)
                     {
                         string meshName = Regex.Replace(s, "BUILTIN/", "");
-                        mesh = Resources.FindObjectsOfTypeAll<Mesh>().Where(mesh => mesh.name == meshName).First();
+                        mesh = Resources.FindObjectsOfTypeAll<Mesh>().Where(m => m.name == meshName).First();
                         Object = new GameObject(meshName);
                         Object.AddComponent<MeshFilter>().sharedMesh = mesh;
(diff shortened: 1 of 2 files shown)


ThomasKerman:
Thanks for this. :)

Josh6680:
You're welcome :)

Issue #103, opened 2015-12-13
Stock asteroid spawner works concurrently with Kopernicus spawner


Starstrider42:
It appears that the stock spawner does not get disabled right away when Kopernicus loads; in fact, on my computer both create a new asteroid almost simultaneously. This bug was observed on a Windows KSP 1.0.5 32-bit install, no mods except Kopernicus 0.6b and the bundled ModuleManager 2.6.13. To reproduce, create a new Sandbox game.

Log available at https://gist.github.com/Starstrider42/3d50125f7ae19df81823. The most relevant entries:
> [AsteroidSpawner]: New object found near Kerbin: Ast. HSJ-227!
>  
> (Filename: C:/buildslave/unity/build/artifacts/StandalonePlayerGenerated/UnityEngineDebug.cpp Line: 56)
> 
> [Kopernicus]: New object found near Kerbin: Ast. OBZ-067!
>  
> ...
>  
> (Filename: C:/buildslave/unity/build/artifacts/StandalonePlayerGenerated/UnityEngineDebug.cpp Line: 56)
> 
> Game State Saved to saves/default/persistent
>  
> (Filename: C:/buildslave/unity/build/artifacts/StandalonePlayerGenerated/UnityEngineDebug.cpp Line: 56)
> 
> [AsteroidSpawner]: No new objects this time. (Odds are 1:2)

ThomasKerman:
I know about that, but currently, I don't see any way to destroy the spawner. The weird thing is, that I destroy the spawner in Awake(), but it seems that he is able to execute his Start() function. ScenarioModules, both, modded and Stock, are applied at the same time, so there should be at least one frame between Awake() and Start().
The only thing I could do would be DestroyImmideate(Resources.FindObjectsOfAll<ScenarioDiscoverableObject>()[0]); but meh.

Starstrider42:
Well, I was hoping if we both attacked the problem from our respective ends we'd make some progress.

In the case of this specific log, the behavior is consistent with a race condition -- if on some computers (mine) Kopernicus is created before ScenarioDiscoverableObjects, then its command to remove the latter might have no effect. You'll notice that the last [AsteroidSpawner] log is right before a scene change, indicating the scenario DOES get deleted on the second load of Kopernicus.

Though I can't explain what I see with Custom Asteroids this way. Unless the return value of `Game.RemoveProtoScenarioModule` doesn't mean what I think it does...

Starstrider42:
[...]

PR #104, opened 2015-12-13
ScenarioDiscoverableObjects now correctly shut down

Starstrider42:
This appears to be a robust fix for #103, with no side effects (e.g., no dangling references to the destroyed scenario).


ThomasKerman:
Thank you!

PR #125, opened 2016-06-02
Fix for bad garbage thrash when panning camera in map/tracking station

Gerry1135:
This significantly improves the horrible garbage thrash when panning the camera around in the map view in OPM.

Further improvements are possible (even desirable) but it will involve rewriting the LoadTexture function to avoid the allocation of a new byte[] containing the entire texture file for each texture that is loaded.

### DIFF
--- a/Kopernicus/Kopernicus.OnDemand/CBAttributeMapSODemand.cs
+++ b/Kopernicus/Kopernicus.OnDemand/CBAttributeMapSODemand.cs
@@ -67,7 +67,7 @@ namespace Kopernicus
                 {
                     CreateMap(Depth, map);
                     IsLoaded = true;
-                    Debug.Log("[OD] CBmap " + name + " enabling self. Path = " + Path);
+                    Debug.Log("[OD] ---> CBmap " + name + " enabling self. Path = " + Path);
                     return;
                 }
 
@@ -90,7 +90,7 @@ namespace Kopernicus
                 IsLoaded = false;
 
                 // Log
-                Debug.Log("[OD] CBmap " + name + " disabling self. Path = " + Path);
+                Debug.Log("[OD] <--- CBmap " + name + " disabling self. Path = " + Path);
             }
 
             // Create a map from a Texture2D
--- a/Kopernicus/Kopernicus.OnDemand/Kopernicus.OnDemand.csproj
+++ b/Kopernicus/Kopernicus.OnDemand/Kopernicus.OnDemand.csproj
@@ -28,6 +28,7 @@
     <DefineConstants>TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Assembly-CSharp">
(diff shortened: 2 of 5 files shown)


CaptRobau:
I'm assuming this fixes this for all mods, since OPM doesn't do anything special map view wise as far as I know.

Gerry1135:
I assume so.  Basically, at the moment, the textures unload as soon as the body goes out of view (it isn't the rings causing the problem, it is the moons of the planet going in and out of the view).  When panning and zooming a body can go out and back in again very quickly causing it to be unloaded and then loaded again very shortly after.  All the fix does is stop the unloading happening immediately.  Instead it waits for a while before unloading in case the body comes back into view (in which case the unload is cancelled).

The delay is currently set to 30 seconds but it could probably be significantly less and still give most of the benefit.

I will be investigating the further optimisation of the LoadTexture function because a similar issue is likely with any mod that loads textures during a scene and it would be very useful to have a less garbage happy way for mods to load them...

Gerry1135:
Actually, the useOnDemand property appears to be true by default so it isn't explicitly specified.  I believe it lives at the Body level so it should be relatively simple to create an MM patch to disable it for all bodies in the meantime (will make it use more memory but some people have plenty and won't mind).  My MM-fu isn't particular strong but I'll try to knock one up later that can be used until this fix (or something like it) gets merged.

Gerry1135:
It appears that the Texture2D class doesn't have any mechanism for loading textures other than from a byte array of the complete file or raw texture data.  In the case of DDS the code actually allocates two byte arrays, one for the initial load which is turned into a MemoryStream, and one for the reading of the raw data from the BinaryReader.  This could be improved by simply passing a FileStream to the BinaryReader rather than reading the whole file into a byte array.

I think a shared "whole file buffer" would be the best that can be done without needing changes in Unity, it really isn't designed for in-scene loading of texture data.  It would result in a fairly large byte array sticking around fairly permanently (it would basically grow to be the size of the largest file that gets loaded and then stick around though it would make sense to set it to a sensibly large initial size to avoid too much reallocation).  I'll add something to this PR later today...

Sigma88:
load on demand can be disabled like this:

```
@Kopernicus:FINAL
{
    %useOnDemand = False
}
```

so it's not tied to any specific body

Gerry1135:
Well, I reckon I'm pretty happy with it now.  There might be a little too much logging but it shouldn't cause a problem and might be useful to diagnose any problems people report.

WazWaz:
What about a MemoryMappedFile for DDS texture loading? That would avoid the buffers.

Gerry1135:
You still have to pass the bulk of the file to the Texture2D.LoadRawTextureData function and that only takes a byte[].  The Texture2D class really needs more useful functions.  You might be able to do something cunning with CreateExternalTexture but you need to pass a native (i.e. DirectX or OpenGL) texture pointer to it which would require platform specific code etc.

Basically, the changes as they are mean that there's only ever one buffer (once it has grown to the size of the largest texture file), this is the best that I've been able to manage so far and any further improvement is likely to be fairly minimal and require some even more involved changes...

ThomasKerman:
Merged this "manually" into development. Thank you very much!

WazWaz:
Ah, I normally work with Unity 5.3, where Texture2D.LoadRawTextureData also has an (IntPtr,size) overload.

Gerry1135:
Does the Unity version of Mono even include MemoryMappedFile?  That's .NET 4 or later and I can't see it in the System.IO I'm looking at...

WazWaz:
Ugh, you're right. Sorry for wasting your time. I forget how ancient it is.

PR #126, opened 2016-06-03
Fix mesh colliders very often being out-of-sync with renderers in KSP 1.1

WazWaz:
It seems in KSP 1.1 they add colliders incrementally in chunks of 64 up to 256, breaking the previous assumptions in Kopernicus.


ThomasKerman:
Merged this "manually" into development. Thank you!

PR #129, opened 2016-06-07
More planetary texture unload delays

Gerry1135:
I've added an unload delay (currently 10 seconds) to MapSO, CBAttributeMapSO and PQSMod_OnDemandHandler.  The first two use a modified version of your old Timer class that auto-starts and has a Kill method to cancel the unload if the texture gets loaded again within the delay.  The other, being derived from MonoBehaviour, uses the same mechanism as the ScaledSpaceDemand class.

It needs more testing as I've only given it a quick go on a Win32 machine using a hugely cut down KSP install running a version of OPM that only includes the Sarnus system and it still crashed after 2 returns to KSC though it was enough for me to Hyperedit a vessel to Ovok orbit and mess around monitoring the log.  I'll release another test version in the forum thread for Poodmund to try and see if it improves his fps drops...


Gerry1135:
I thought I saw all three of the new delayed unload functions being called when I tested on Win32 earlier but I'm not seeing them on Win64 with either 32 or 64 bit KSP so I'm not convinced it's going to make any difference.  Also, the PQSMod_OnDemandHandler calls the other two so they don't all need the delay anyway.  I'll make some more changes tomorrow, possibly just removing the new delays I've added but leaving the new configuration setting in place for ScaledSpaceDemand.

ThomasKerman:
Totally forgot this PR! Merging!

PR #132, opened 2016-06-21
Add new navballSwitchRadiusMultLow to Properties

NathanKell:
Note: 113 also fixes navballSwitchRadiusMult not being used at all ever...


Sigma88:
what does this do? (if you don't mind me asking)

NathanKell:
KSP defaults to swapping to ORBIT mode on the navball at 6% of the body radius in altitude (36km in stock), and 5.5% on the way down (swapping to SURF). 1.0.5 made that customizable, but only the 6% one. It was borked in 1.1. 1.1.3 fixes that, and adds customizing the "on the way down" one too.

PR #135, opened 2016-07-03
Orbit iconColor

Sigma88:
I thought this could have been an interesting addition.
I'm not sure you want to add the option into kopernicus, but I thought to mention it :D


ThomasKerman:
Why not.

I have no idea why I notice this *now*

ThomasKerman:
Btw, funny that I can merge this even though github says "from unknown repository"

PR #143, opened 2016-11-04
RDVisibility.HIDDEN now hides only the tagged body

Sigma88:
entries for the children are moved to where the hidden body should be

alternatively, you could add a new option to RDVisibility, for example RDVisibility.SKIP

so that you can keep both option:
HIDDEN > to hide the body and all its children
SKIP > to hide that body only

### DIFF
--- a/Kopernicus/Kopernicus/Configuration/PropertiesLoader.cs
+++ b/Kopernicus/Kopernicus/Configuration/PropertiesLoader.cs
@@ -284,7 +284,8 @@ namespace Kopernicus
             {
                 VISIBLE,
                 NOICON,
-                HIDDEN
+                HIDDEN,
+                SKIP
             }
 
             // Apply Event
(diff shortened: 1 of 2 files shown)


ThomasKerman:
Could you add code to undo the changes to the PSystem after rebuilding the planets? I think that would be cleaner.

Sigma88:
sure

Sigma88:
I decided to go with the "SKIP" option since that helps simplify the code to undo the changes to the psystem

When I have some time I'll add the bit of code to account for PostSpawnOrbit and then you can merge it

(or at least take another look at it to decide if it's fine)

Sigma88:
also, I added SKIP as the last option
( RDVisibility = 3 )
because I didn't want to disrupt the current system

but probably it would make more sense to have it as second-to-last

0 - VISIBLE
1 - NOICON
2- SKIP         (hide selected body only)
3- HIDDNE   (hide body and all its children)

Sigma88:
ok, I think it should be fine now. try giving it a look whenever you have some time ;)

Sigma88:
fixed the way I get the method (now it uses the name)
🙉

PR #144, opened 2016-11-06
ZoomFixer

Sigma88:
this new parameter will work by defining the amount of meters that can fit vertically into the screen

let's say you want the zoom to stop when Jool is taking half of the vertical space of the screen

since jool diameter is 12,000,000 meters you need to set

Body/Properties/**maxZoom = 24000000**


Sigma88:
moved inside FixZooming() since you had already Zoom-related code there

Sigma88:
ok, now it works fine it would seem.
do we really need 
|| MapView.MapIsEnabled

because I tried without and it works the same, even in map view

PR #146, opened 2016-11-11
custom game clock

Sigma88:
This has different advantages over the stock clock

Mainly I went out of my way to make sure the game clock was consistent with the sun position in the sky


Sigma88:
nice :)

PR #148, opened 2016-11-17
Custom clock

Sigma88:
some improvements on useKopernicusTime

I forgot to actually check if useKopernicusTime was true before applying the changes, that's fixed now.

As someone pointed out (can't remember who) if some numbers were not good the code reverted to Kerbin time. Now it reverts to the time chosen by the user (Kerbin or Earth).

I also added the option of completely customize the clock.
For each unit of time you can define:
- **singular and plural name** - replacing the standard = *Second(s), Min(s), Hour(s), Day(s), Year(s)*
- **symbol** - replacing the standard = *s, m, h, d, y*
- **value** - replacing the standard = *1, 60, 3600, 21600/86400, 9201600/31536000*

***"value" has to be defined in real life seconds, spanning the whole duration of the unit of time***

I didn't put any checks to make sure the numbers used are valid or make sense (seconds should be shorter than minutes, and minutes shorter than hours) maybe I should add that.

The custom clock is defined as follows:
```
Kopernicus
{
    KopernicusTime
    {
        Second
        {
            singular = [string]
            plural = [string]
            symbol = [string]
            value = [double]
        }
        Minute
        {
            singular = [string]
            plural = [string]
            symbol = [string]
            value = [double]
        }
        Hour
        {
            singular = [string]
            plural = [string]
            symbol = [string]
            value = [double]
        }
        Day
        {
            singular = [string]
            plural = [string]
            symbol = [string]
            value = [double]
        }
        Year
        {
            singular = [string]
            plural = [string]
            symbol = [string]
            value = [double]
        }
    }
}
```


Sigma88:
wrong place

PR #149, opened 2016-11-17
Custom clock

Sigma88:
some improvements on useKopernicusTime

I forgot to actually check if useKopernicusTime was true before applying the changes, that's fixed now.

As someone pointed out (can't remember who) if some numbers were not good the code reverted to Kerbin time. Now it reverts to the time chosen by the user (Kerbin or Earth).

I also added the option of completely customize the clock.
For each unit of time you can define:
- **singular and plural name** - replacing the standard = *Second(s), Min(s), Hour(s), Day(s), Year(s)*
- **symbol** - replacing the standard = *s, m, h, d, y*
- **value** - replacing the standard = *1, 60, 3600, 21600/86400, 9201600/31536000*

***"value" has to be defined in real life seconds, spanning the whole duration of the unit of time***

I didn't put any checks to make sure the numbers used are valid or make sense (seconds should be shorter than minutes, and minutes shorter than hours) maybe I should add that.

The custom clock is off by default, it get turned on when at least one of these conditions are met:
1- **Kopernicus/useKerbinTime = true**
2- **Kopernicus/KopernicusTime {}**    *exists (even if empty)*

Syntax to define the clock:
```
Kopernicus
{
    KopernicusTime
    {
        Second
        {
            singular = [string]
            plural = [string]
            symbol = [string]
            value = [double]
        }
        Minute
        {
            singular = [string]
            plural = [string]
            symbol = [string]
            value = [double]
        }
        Hour
        {
            singular = [string]
            plural = [string]
            symbol = [string]
            value = [double]
        }
        Day
        {
            singular = [string]
            plural = [string]
            symbol = [string]
            value = [double]
        }
        Year
        {
            singular = [string]
            plural = [string]
            symbol = [string]
            value = [double]
        }
    }
}
```


Olympic1:
It was me who pointed out the revert to kerbin time. Hope I didn't cause you too much trouble with this.

Sigma88:
@Olympic1  no truble at all, it was just 2 lines, most of the work was to make the custom clock.
I actually managed to make it work on my first try, which is something never happens.

I'm testing the dlls and honestly I can't find issues with it

Sigma88:
added some bonus features:

1- now I check that the definitions make sense (+inf > hour > minute > second > 0)
if they don't follow that pattern the custom clock doesn't get used

2- smarter use of integers to allow the clock to tick faster/slower instead of once every IRL second

Sigma88:
🐒🐒🐒

ThomasKerman:
Looks good

infoman:
So, to activate this, all I need is to create a config file with something like:

```
Kopernicus
{
    useKopernicusTime = true
    KopernicusTime
    {}
}
```

?

ThomasKerman:
You dont even need the empty KopernicusTime node

infoman:
Apparently it doesn't work:
```
@Kopernicus:FOR[Kopernicus]
{
    %useKopernicusTime = true
}
```

Rotation period

Exception in Player.log:
```
NullReferenceException: Object reference not set to an instance of an object
  at Kopernicus.ClockFixer.Start () [0x00000] in <filename unknown>:0 
```

[Full log](https://github.com/Kopernicus/Kopernicus/files/671802/Player.log.txt)

PR #165, opened 2017-04-04
Ring shader bug fix

LGhassen:
Hello,

This fixes the issue with the new ring shader not appearing on some bodies. I also added the shader compilation cache and log files to the .gitignore

-blackrack


Poodmund:
The issue was that when the new ring shader was used, it didn't display properly when used on more than one body.

The above commits seem to fix this issue.

LGhassen:
Actually Pood, from my testing the other day, the number of actual bodies using rings didn't seem to matter, specific bodies get it anyway, even if their the only one with rings. Anyway, if it's fixed, it's fixed.

StollD:
Thank you @LGhassen 

I am going to merge this manually, pull requests should go against development (/me slaps himself for not adding a CONTRIBUTING.md), and they shouldnt contain compiled binaries, since those are built by Jenkins.
So more my fault than yours. ;)

StollD:
Merged manually

Issue #183, opened 2017-06-16
Kopernicus not setting biome displayname


DMagic1:
In KSP 1.3 biomes now have a localized name field.

[CBAttributeMapSO.MapAttribute.displayName](https://kerbalspaceprogram.com/api/class_c_b_attribute_map_s_o_1_1_map_attribute.html#ac0c5207b4a1257f963f805274a094a47)

This is read for science results among other things.

The BiomeLoader class doesn't appear to handle this, only setting the regular name field.

An option to set a display name would be good for localization, but it should at least set the display name using the regular name field by default.

StollD:
Done by @Sigma88

PR #184, opened 2017-06-17
attempt to solve #183

Sigma88:


Kopernicus/Kopernicus/Configuration/PropertiesLoader.cs
```diff
@@ -209,7 +209,7 @@ public NumericParser<double> navballSwitchRadiusMultLow
 
             // Biomes of this body
             [PreApply]
-            [ParserTargetCollection("Biomes", nameSignificance = NameSignificance.None)]
+            [ParserTargetCollection("Biomes", nameSignificance = NameSignificance.Key, key = "Biome")]
```
StollD: Please revert that. It isn't actually needed, but it will break those mods who use different names

Issue #190, opened 2017-07-04
Cannot use BUILTIN map in more than one PQS mod: "Operation is not valid due to the current state of the object"


gordonfpanam:
When I try to use the same built-in height map for more than one PQS object on a given world, Kopernicus will throw an exception when attempting to use that map more than once.

```
Body
{
   PQS
   {
      Mods
      {
         LandControl
         {
            heightMap = BUILTIN/[example_height]
         }
         VertexHeightMap
         {
            map = BUILTIN/[example_height]
         }
      }
   }
   Ocean
   {
      Mods
      {
         RemoveQuadMap
         {
            map = BUILTIN/[example_height]
         }
      }
   }
}
```
Doing this will cause an exception when Kopernicus encounters the second use of BUILTIN/[example_height] here. A more complete example is available in my [Alien Space Programs](https://github.com/gordonfpanam/Alien-Space-Programs) fork.

I tested this with BUILTIN/kerbin_height_new and with BUILTIN/desertplanet_height at least, and get the same behaviour. This behaviour does not occur with external images. It also doesn't appear to occur with ScaledVersion { Material {} }, where I can use things like BUILTIN/KerbinScaledSpace300 for both texture and mainTex settings, and BUILTIN/KerbinScaledSpace401 for both normals and bumpMap settings.

It also doesn't matter when the second usage occurs; I could put VertexHeightMap before LandControl and get the exception to occur in the LandControl section. I could substitute an external file at any point and avoid the exception, as long as the BUILTIN object is used only once.

I wouldn't be trying such a thing, except this is what kittopia-tech spits out when I save a stock world's configuration for use later. I'm also trying to make a Kerbin clone without the Kerbin template to avoid issue #186. Using the built-in maps would save a lot of memory in-game, and disk space overall, when trying to clone the stock worlds.

I tested this as far back as Kopernicus 1.2.2-1, and every iteration up to 1.2.2-9 on KSP 1.2.2 and 1.3.0-4 on KSP 1.3.

Sigma88:
@gordonfpanam could you try this version of Kopernicus and let me know if it solves your issue?

[KopernicusTESTSolve190.zip](https://github.com/Kopernicus/Kopernicus/files/1121985/KopernicusTESTSolve190.zip)

gordonfpanam:
This is definitely an improvement. I can access BUILTIN/kerbin_height_new and BUILTIN/desertplanet_height more than once now.

[...]

PR #191, opened 2017-07-04
Don't Rename Stock Textures

Sigma88:
this solves #190 

@StollD I can't see any reasons why the lines I deleted were there, so if you know why they are needed and this PR could break something else, feel free to close it

### DIFF
--- a/Kopernicus/Kopernicus/Configuration/Parser/BuiltinTypeParsers.cs
+++ b/Kopernicus/Kopernicus/Configuration/Parser/BuiltinTypeParsers.cs
@@ -75,12 +75,12 @@ namespace Kopernicus
         public class Texture2DParser : IParsable
         {
             public Texture2D value;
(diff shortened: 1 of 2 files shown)


Kopernicus/Kopernicus/Configuration/Parser/BuiltinTypeParsers.cs
```diff
@@ -142,7 +142,7 @@ public void SetFromString(string s)
 
                 if (s.StartsWith("BUILTIN/"))
                 {
-                    s.Replace("BUILTIN/", "");
+                    s = s.Length == 8 ? "" : s.Substring(8);
```
StollD: Umm, this is the same as `s = s.Substring(8);`, since s.Substring(s.Length) returns "".
And it cannot be less than 8 chars

PR #199, opened 2017-07-30
Test

Sigma88:


Sigma88:
for some reason this solves the nullref exceptions

but basically it reverts all stuff you (@stolld) added between 1.3.0-4 and now.

PR #201, opened 2017-08-03
Add properties for a thick ring

HebaruSan:
This patch extends Ring and RingLoader:

- Support thickness, i.e. a distance between the top and bottom face of the ring, with inner and outer faces added in between
- Support for tiling the texture onto the various faces of the resulting cylinder
- Support for giving each ring its own rotation period

The changes are designed to be backwards compatible; existing planets will not be changed unless they are modified to support the new features (tested with JoolRings from KopernicusExamples). Also hoped to be future compatible with features like ring-on-ring shadows and a solid landable terrain surface.

Intended for visualizing classic fictional megastructures, for example:

ringworld_screenshot

(I'll be releasing the config for that screenshot as a separate mod if this patch is merged.)

### DIFF
--- a/Kopernicus/Kopernicus.Components/Ring.cs
+++ b/Kopernicus/Kopernicus.Components/Ring.cs
@@ -4,7 +4,7 @@
  * Created by: BryceSchroeder and Teknoman117 (aka. Nathaniel R. Lewis)
  * Maintained by: Thomas P., NathanKell and KillAshley
  * Additional Content by: Gravitasi, aftokino, KCreator, Padishar, Kragrathea, OvenProofMars, zengei, MrHappyFace, Sigma88
- * ------------------------------------------------------------- 
+ * -------------------------------------------------------------
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
  * License as published by the Free Software Foundation; either
@@ -19,11 +19,11 @@
  * License along with this library; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
  * MA 02110-1301  USA
- * 
+ *
  * This library is intended to be used as a plugin for Kerbal Space Program
  * which is copyright 2011-2015 Squad. Your usage of Kerbal Space Program
  * itself is governed by the terms of its EULA, not the license above.
- * 
+ *
  * https://kerbalspaceprogram.com
  */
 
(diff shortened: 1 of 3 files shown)


HebaruSan:
These texture tiling changes also allow traditional rings to add "horizontal" detailing, since the full texture is now applied as an area rather than just one linear sample; here's JoolRings after a "tiles = 5" line is added:

jool-weird-rings

HebaruSan:
Added per-ring rotation period configuration in latest commit to support rotation speeds other than 0 and parent-matching.

HebaruSan:
@PhineasFreak , how do you add inclination to a ring? Is it the `angle` property?

Poodmund:
Yes, the angle property denotes the inclination of the ring in degrees, 0 being perpendicular to the planets rotational axis.

HebaruSan:
OK, I think that's relatively easy to do; see below for some thick tiled JoolRings with rotation and inclination. Can anyone please provide a confirmed-broken config that I can test to make sure it's working as desired?

rings_of_uranus

HebaruSan:
On second look, the changes I've made aren't very different from what the old code was doing for lockRotation=true, so they probably don't fix this issue. I'll need good instructions for how to reproduce this issue if you want it fixed.

HebaruSan:
@PhineasFreak , are you sure about the problem with the LAN? I found this post on the forum which seems to suggest that inclined rings have already been done this way, and that the problem isn't that the LAN can change, but that you can't change it in the Kopernicus cfg files. Adding a setting for LAN would be straightforward, if that's what's needed.

http://forum.kerbalspaceprogram.com/index.php?/topic/140580-130-122-kopernicus-release-4-june-15/&do=findComment&comment=3130510

PhineasFreak:
@HebaruSan I may be wrong about the LAN but i did find a post by the user **OhioBob** that describes the problem exactly:

http://forum.kerbalspaceprogram.com/index.php?/topic/140580-130-122-kopernicus-release-4-june-15/&do=findComment&comment=2805897

It is an old post but i am not sure if anyone ever took a look at it.

My idea was that the ring would stay locked so that it will never end up being half in the dark and half in the light side of the body (much like how ISS always has the cupola facing the surface of the Earth).

I will post some reference screenshots of Uranus later when i will have access to my workstation.

HebaruSan:
Thanks for that link! I was able to observe the issue with the JoolRings example by changing Laythe's inclination and using its orbit as a reference plane. Indeed, the ring's LAN does jump around from one restart to the next, and I now have the screenshots to prove it. (Of course, figuring out why is another question entirely...)

HebaruSan:
A fix for the LAN issue is now available in #204, and I've merged it into this PR as well to avoid future conflicts. (I also deleted some of my replies here that were no longer useful.)

PR #206, opened 2017-08-19
Ring shader fixes and enhancements

HebaruSan:
This patch makes the configuration of the "new shader" more consistent with the old shader, so old rings can be updated to use it without modification:

- Use mesh's texture coordinates (fixes the problem where you have to rotate the textures to use the new shader, so the shaders can now share textures)
- Obey `color` property (required for tinted rings as in the green JoolRings example)
- Plus a few quality of life improvements like z-buffering and backface culling

Then a new concept of "inner shade" is added to the "thick" rings. This is a texture that can be used to cast independently mobile shadows on the inner face of a thick ring.

- Property for the texture
- Property specifying how many times it should be tiled
- Property for its rotation period

After these changes, I'm able to configure a real-scale Ringworld complete with functional shadow squares:

newshader_and_shadowsquareshadows

### DIFF
--- a/Kopernicus/Kopernicus.Components/Ring.cs
+++ b/Kopernicus/Kopernicus.Components/Ring.cs
@@ -152,6 +152,7 @@ namespace Kopernicus
                 {
                     ringMR.material.SetFloat("planetRadius",       planetRadius);
                     ringMR.material.SetFloat("penumbraMultiplier", penumbraMultiplier);
+                    ringMR.material.SetFloat("sunRadius",          (float) KopernicusStar.Current.sun.Radius);
                 }
 
                 ringMR.material.color = color;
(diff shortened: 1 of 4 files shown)


StollD:
Looks like I didn't see this. Looks good, thank you!

HebaruSan:
Cool!
Just a quick note on this one, it does require the assets in [Distribution/.../Shaders](https://github.com/Kopernicus/Kopernicus/tree/master/Distribution/Release/GameData/Kopernicus/Shaders) to be rebuilt using the [shader-export](https://github.com/Kopernicus/shader-export) project. I was not sure whether to include that in my commits or not; it might be considered to be part of packing a release. Either way, let me know; if you'd like new versions of these files, I can update them.

StollD:
Sure, feel free to update them. I will probably just forget it when I make the next release.
          
            
            On Aug. 27 2017, at 8:54 pm, HebaruSan <notifications@github.com> wrote:
            
            
              Cool!
Just a quick note on this one, it does require the assets in Distribution/.../Shaders to be rebuilt using the shader-export project. I was not sure whether to include that in my commits or not; it might be considered to be part of packing a release. Either way, let me know; if you'd like new versions of these files, I can update them.

HebaruSan:
Sounds good, I submitted them in:
https://github.com/Kopernicus/Kopernicus/pull/210

PR #221, opened 2017-09-07
make sure the PSO_referencebody exists

Sigma88:
this adds the same exception from #220 to PostSpawnOrbits

### DIFF
--- a/Kopernicus/Kopernicus/RuntimeUtility/RuntimeUtility.cs
+++ b/Kopernicus/Kopernicus/RuntimeUtility/RuntimeUtility.cs
@@ -156,9 +156,16 @@ namespace Kopernicus
 
                     CelestialBody newRef = PSystemManager.Instance.localBodies.FirstOrDefault(b => b.transform.name == loader.referenceBody);
                     if (newRef != null)
+                    {
                         body.orbit.referenceBody = body.orbitDriver.referenceBody = newRef;
                     else
-                        throw new Exception("PostSpawnOrbit reference body for \"" + body.name + "\" could not be found. Missing body name is \"" + loader.referenceBody + "\".");
+                    {
+                        // Log the exception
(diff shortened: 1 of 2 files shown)


Kopernicus/Kopernicus/RuntimeUtility/RuntimeUtility.cs
```diff
@@ -153,6 +153,13 @@ void Start()
                     body.orbitDriver.orbit = loader.orbit;
                     CelestialBody oldRef = body.referenceBody;
                     body.referenceBody.orbitingBodies.Remove(body);
+
+                    CelestialBody newRef = PSystemManager.Instance.localBodies.FirstOrDefault(b => b.transform.name == loader.referenceBody);
+                    if (newRef != null)
+                        body.orbit.referenceBody = body.orbitDriver.referenceBody = newRef;
+                    else
+                        throw new Exception("PostSpawnOrbit reference body for \"" + body.name + "\" could not be found. Missing body name is \"" + loader.referenceBody + "\".");
```
StollD: Please make this a `Debug.Log`. The system loaded successfully, no need to kill the runtime maintainance when the system could work.

StollD: That said, we should add a check that disables RuntimeUtility if the loader failed.

Sigma88: not sure about "could work"

I mean, if it's designed to work with the PostSpawnOrbit and an user messes up when updating the planet pack to a newer version, it could result in a save loading on a system different than what it is supposed to be, potentially losing crafts and crybabies going nuts.

that said, I can change it if you are convinced Debub.Log will be a better choice

StollD: Make it a Debug.Log and spawn the "something broke search help" window?

Sigma88:
ok, I think this should be fine now

PR #223, opened 2017-09-07
fix cachefile generation

Sigma88:
correctly account for the presence of a non-spherical ocean when generating the cachefile for the planet


Sigma88:
I've tested this extensively and couldn't find any issues yet

PR #230, opened 2017-09-21
Sigma 1

Sigma88:
added some stuff to enhance compatibility

### DIFF
--- a/Kopernicus/Kopernicus/Configuration/Loader.cs
+++ b/Kopernicus/Kopernicus/Configuration/Loader.cs
@@ -209,6 +209,9 @@ namespace Kopernicus
                     Logger.Default.SetAsActive ();
                 }
 
+                // Event
+                Events.OnLoaderLoadedAllBodies.Fire(this, node);
+
                 // Load all of the asteroids                
                 foreach (ConfigNode asteroidNode in node.GetNodes(asteroidNodeName))
                 {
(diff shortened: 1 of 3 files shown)


Sigma88:
these changes are simple enough that I think there shouldn't be any problems in adding them

@StollD  maybe take a look to make sure I didn't do something stupid :D

StollD:
I don't know if I like that currentBody stuff to be honest...

If you really want it, please add currentBody to BaseLoader, make it default to generatedBody, and change all the loaders to include a third constructor that accepts a PSystemBody. (which then sets currentBody)

Sigma88:
gotcha

Sigma88:
can I change the generatedBody to be non-static?

Sigma88:
I think this should do the trick

        public class BaseLoader
        {
            public PSystemBody currentBody = null;

            // Singleton of the currently edited body
            public PSystemBody generatedBody
            {
                get
                {
                    if (currentBody != null)
                        return currentBody;
                    else
                        return Loader.currentBody.generatedBody;
                }
            }
        }

Sigma88:
alternatively we could use


            public PSystemBody currentBody { internal get; set; }

instead of 

            public PSystemBody currentBody = null;

StollD:
I like your first suggestion, but I would suggest to make both, generatedBody and currentBody protected instead of public

PR #233, opened 2017-09-22
Starfixes

Sigma88:
some fixes for stars in the RnD

### DIFF
--- a/Kopernicus/Kopernicus/Configuration/PropertiesLoader.cs
+++ b/Kopernicus/Kopernicus/Configuration/PropertiesLoader.cs
@@ -40,7 +40,7 @@ namespace Kopernicus
 
             // Body description
             [ParserTarget("description")]
-            public String description 
+            public String description
             {
                 get { return celestialBody.bodyDescription; }
                 set { celestialBody.bodyDescription = value; }
@@ -48,20 +48,20 @@ namespace Kopernicus
 
             // Radius
             [ParserTarget("radius")]
-            public NumericParser<Double> radius 
+            public NumericParser<Double> radius
             {
                 get { return celestialBody.Radius; }
                 set { celestialBody.Radius = value; }
             }
-            
+
             // GeeASL
             [ParserTarget("geeASL")]
-            public NumericParser<Double> geeASL 
+            public NumericParser<Double> geeASL
             {
                 get { return celestialBody.GeeASL; }
                 set { celestialBody.GeeASL = value; }
             }
-            
+
             // Mass
             [ParserTarget("mass")]
             public NumericParser<Double> mass
@@ -79,7 +79,7 @@ namespace Kopernicus
                 set { celestialBody.gMagnitudeAtCenter = celestialBody.gravParameter = value; hasGravParam = true; }
             }
             private Boolean hasGravParam = false;
-            
+
             // Does the body rotate?
             [ParserTarget("rotates")]
             public NumericParser<Boolean> rotates
@@ -87,7 +87,7 @@ namespace Kopernicus
                 get { return celestialBody.rotates; }
                 set { celestialBody.rotates = value; }
             }
-            
+
             // Rotation period of the world
             [ParserTarget("rotationPeriod")]
             public NumericParser<Double> rotationPeriod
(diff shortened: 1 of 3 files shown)


Sigma88:
this could be considered a bug fix, but I will let you merge it @StollD because I added something that might be considered a new feature:

killing star rotation in the RnD

if you want we could add a parameter like ```Properties/RDRotation``` (true by default for all bodies except for stars) so that the list of planets for which the rotation is killed can be edited by the user

Kopernicus/Kopernicus/RuntimeUtility/RnDFixer.cs
```diff
@@ -40,6 +40,18 @@ namespace Kopernicus
     /// </summary>
     public class RnDFixer : MonoBehaviour
     {
+        List<RDPlanetListItemContainer> stars = new List<RDPlanetListItemContainer>();
```
StollD: Please add an explicit `private` and move the initialization to `Start()`

Kopernicus/Kopernicus/RuntimeUtility/RnDFixer.cs
```diff
@@ -40,6 +40,18 @@ namespace Kopernicus
...
+
+        void LateUpdate()
+        {
+            // Block the orientation of all stars
+            for (int i = 0; i < stars?.Count(); i++)
```
StollD: `stars` is never null so you can remove the `?` (I am suprised that this compiles, since you are comparing an int to null)

Kopernicus/Kopernicus/RuntimeUtility/RnDFixer.cs
```diff
@@ -40,6 +40,18 @@ namespace Kopernicus
...
+            {
+                RDPlanetListItemContainer star = stars.ElementAt(i);
```
StollD: `stars[i]` - No need to use Linq and waste memory.

Kopernicus/Kopernicus/RuntimeUtility/StarLightSwitcher.cs
```diff
@@ -50,8 +50,8 @@ static StarComponent()
         void Start()
         {
             // Find the celestial body we are attached to
-            celestialBody = PSystemManager.Instance.localBodies.First(body => body.scaledBody == gameObject);
-            Logger.Default.Log("StarLightSwitcher.Start() => " + celestialBody.bodyName);
+            celestialBody = PSystemManager.Instance.localBodies.FirstOrDefault(body => body.scaledBody == gameObject);
```
StollD: While I agree that this shouldn't throw the message it currently throws, it should throw something. Maybe something like this

```csharp
if (celestialBody == null)
    throw new InvalidOperationException("The StarComponent isn't attached to the ScaledSpace object of a body.");
```

Kopernicus/Kopernicus/RuntimeUtility/RnDFixer.cs
Sigma88: an int cannot be lower than a null so it returns false

I had an issue once where a null list was breaking my code, since then I've always used ```?``` to make sure that even if I have a null the .Count() method doesn't trigger a nullref

is there a downside to using the ```?``` ?

StollD: It is extra code that gets executed every frame and that isn't neccessary.

Kopernicus/Kopernicus/RuntimeUtility/StarLightSwitcher.cs
Sigma88: the problem is that in the RnD all stars have a StarComponent

I don't think they should thrown an error everytime you enter the RnD

StollD: Ok, then leave it that way

Sigma88:
ok, I think I've addressed everything

you are still using .ElementAt in the RnDFixer.cs 
do you want me to change that as well?

Kopernicus/Kopernicus/Configuration/PropertiesLoader.cs
```diff
@@ -264,8 +264,21 @@ public NumericParser<Boolean> selectable
                 set { if (!value.value) celestialBody.Set("notSelectable", true); }
             }
 
-            // If the body should be hidden in RnD
+            // If the body should be hidden in RD
             [ParserTarget("RDVisibility")]
+            public EnumParser<RDVisibility> hiddenRD
```
StollD: You can just add two ParserTargets to one property ;)

StollD:
Yes, that would be cool

PR #234, opened 2017-09-22
add customizeable option RDRotation

Sigma88:


Sigma88:
if you also want the customizeable option merge this PR first, and then merge #233

Kopernicus/Kopernicus/Configuration/PropertiesLoader.cs
```diff
@@ -277,6 +277,19 @@ public EnumParser<RDVisibility> hiddenRnD
                 set { celestialBody.Set("hiddenRnD", value.value); }
             }
 
+            // If the body should be hidden in RnD
+            [ParserTarget("RDRotation")]
+            public NumericParser<bool> RDRotation
```
StollD: Can you change this to `RnDRotation`?

Kopernicus/Kopernicus/RuntimeUtility/RnDFixer.cs
```diff
@@ -40,14 +40,14 @@ namespace Kopernicus
     /// </summary>
     public class RnDFixer : MonoBehaviour
     {
-        List<RDPlanetListItemContainer> stars = new List<RDPlanetListItemContainer>();
+        internal static List<RDPlanetListItemContainer> RDRotationKill = new List<RDPlanetListItemContainer>();
```
StollD: `RnDRotationKill`?

Kopernicus/Kopernicus/RuntimeUtility/RnDFixer.cs
```diff
@@ -252,9 +252,10 @@ void Start()
                     }
                 }
 
-                if (planetItem?.planet?.GetComponent<StarComponent>() != null)
+                // Kill Rotation
+                if (body.Has("RDRotation") ? !body.Get<bool>("RDRotation") : planetItem?.planet?.GetComponent<StarComponent>() != null)
```
StollD: Tbh, I would just check if the body is in KopernicusStar.Stars, but this works too.

Kopernicus/Kopernicus/Configuration/PropertiesLoader.cs
```diff
@@ -277,6 +277,19 @@ public EnumParser<RDVisibility> hiddenRnD
...
+            {
+                get
+                {
+                    if (celestialBody.Has("RDRotation"))
+                        return generatedBody.Get<bool>("RDRotation");
+                    return generatedBody?.scaledVersion?.GetComponentsInChildren<SunShaderController>(true)?.Length > 0;
+                }
+                set { celestialBody.Set("RnDrotation", value.value); }
```
StollD: please fix this.

Kopernicus/Kopernicus/Configuration/PropertiesLoader.cs
Sigma88: should we change **RD**Visibility as well?

or do you want to have **RD**Visibility  and **RnD**Rotation

Kopernicus/Kopernicus/RuntimeUtility/RnDFixer.cs
Sigma88: it's the list of bodies for which we want to "kill" the rotation

it's not required that they are stars, but someone might want to kill the rotation of other bodies

I initially had named it stars  because I didn't intend to allow the users to choose which bodies to select

Kopernicus/Kopernicus/RuntimeUtility/RnDFixer.cs
Sigma88: I wasn't sure how KopernicusStar worked

do you keep all stars in there? even those with "GivesOffLight = False" ?

StollD: Yes

Kopernicus/Kopernicus/RuntimeUtility/RnDFixer.cs
StollD: I know. I was just suggesting a new name (with an n)

Kopernicus/Kopernicus/RuntimeUtility/RnDFixer.cs
Sigma88: honestly I would restrict the "off" feature only to those stars that have coronas, since the only downside of rotating stars is that the corona rotates together with the star >_>

Kopernicus/Kopernicus/Configuration/PropertiesLoader.cs
StollD: Seems like a good idea, but I wouldn't change it. Instead we should add a second ParserTarget with the new name to it.

Kopernicus/Kopernicus/RuntimeUtility/RnDFixer.cs
StollD: Thats fine.

PR #235, opened 2017-09-23
Rd fix

Sigma88:
not the most beautiful fix, but it works so I don't complain


Sigma88:
@StollD  after you merge this, feel free to release Kopernicus

I couldn't find any more bugs or features the new release might need

PR #238, opened 2017-09-28
Displayname

Sigma88:


src/Kopernicus/Configuration/BiomeLoader.cs
```diff
@@ -37,35 +38,35 @@ public class BiomeLoader : IParserEventSubscriber
...
             }
 
             // The displayName of this biome
             [ParserTarget("displayName")]
-            public String displayName 
+            public String displayName
             {
-                get { return attribute.displayname; } // This is not displayName because of reasons
-                set { attribute.displayname = value; }
+                get { return Localizer.Format(attribute.localizationTag); } // This is not displayName because of reasons
```
StollD: Could we just return the value without passing it to Localizer.Format?

src/Kopernicus/Configuration/PropertiesLoader.cs
```diff
@@ -338,6 +337,10 @@ void IParserEventSubscriber.Apply(ConfigNode node)
             // PostApply Event
             void IParserEventSubscriber.PostApply(ConfigNode node)
             {
+                // Replace biomes
+                if (biomes?.Count() > 0)
+                    celestialBody.BiomeMap.Attributes = biomes.Select(b => b.attribute).ToArray();
```
StollD: Can you add a check if the biome map is null, and if yes, throw an appropriate exception?

PR #251, opened 2017-12-11
Ram and OnDemand

Sigma88:
this fixes a couple of issues:

1- CBAttributeMapSODemand.Unload()

this method wasn't working properly since it was expecting textures to have an empty path, now it's fixed

2- Utility.ComputeScaledSpaceMesh()

this method was using a lot of RAM because it was calling OnDemand.OnDemandStorage.EnableBody() but it wasn't calling OnDemand.OnDemandStorage.DisableBody(), now this is fixed as well

### DIFF
--- a/src/Kopernicus.OnDemand/CBAttributeMapSODemand.cs
+++ b/src/Kopernicus.OnDemand/CBAttributeMapSODemand.cs
@@ -81,7 +81,7 @@ namespace Kopernicus
             public void Unload()
             {
                 // We can only destroy the map, if it is loaded and initialized
-                if (!IsLoaded || !String.IsNullOrEmpty(Path))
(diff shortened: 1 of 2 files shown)


src/Kopernicus.OnDemand/CBAttributeMapSODemand.cs
```diff
@@ -81,7 +81,7 @@ public void Load()
             public void Unload()
             {
                 // We can only destroy the map, if it is loaded and initialized
-                if (!IsLoaded || !String.IsNullOrEmpty(Path))
+                if (!IsLoaded || String.IsNullOrEmpty(Path))
```
StollD: It is funny how nobody noticed this for now almost 3 years :D

PR #269, opened 2018-02-15
Fix245

Sigma88:
maybe it would be enough to use
```
if (Current != null)
```

not sure which of the two would be better


StollD:
```
if (Current != null && Current.lensFlare != null)
```

Sigma88:
done

PR #275, opened 2018-02-20
fixes for the main menu

Sigma88:


StollD:
What I don't understand is: Why do we need another dummy component if we can do the same work in RuntimeUtility.Awake()/Start()? Having another component just bloats the code.

If you want to keep it please move it to it's own file at least.

Also, that KopernicusMainMenu flag should be in Templates I think.

Sigma88:
[RuntimeUtility runs only once](https://github.com/Kopernicus/Kopernicus/blob/94f2cd87fab328d1ab925070f5f9b824b2a1b34c/src/Kopernicus/RuntimeUtility/RuntimeUtility.cs#L44) but we need the code for the mainmenu to run at every "MainMenu.Awake() / .Start()"  (that's why you had it in "Update()" before, which is why it was leaving one frame of un-edited mainmenu in the first place)

that is also why I made a [new component](https://github.com/Kopernicus/Kopernicus/blob/94f2cd87fab328d1ab925070f5f9b824b2a1b34c/src/Kopernicus/RuntimeUtility/RuntimeUtility.cs#L555) if there is a better place to put the code just let me know

Sigma88:
also, is there a way to have the boolean setup so that it doesn't require external plugins to have Kopernicus as a reference?

Sigma88:
if you prefere leaving it the way it was working before it's fine for me

as long as there is a switch to disable the kopernicus main menu editor, I'm fine with anything else you want to do

Sigma88:
I've written this alternative as well

https://github.com/Kopernicus/Kopernicus/tree/MenuAlternative

I don't know how stable this would be since I've only tested it quickly

PR #302, opened 2018-07-31
Duplicate body/asteroid loader fix

Corfiot:
Duplicate bodies/asteroids cause duplicate loggers to be created, throwing IOExceptions and killing loading. Now loader skips duplicated entries gracefully.

### DIFF
--- a/src/Kopernicus/Configuration/Loader.cs
+++ b/src/Kopernicus/Configuration/Loader.cs
@@ -206,6 +206,8 @@ namespace Kopernicus
                 // Dictionary of bodies generated
                 Dictionary<String, Body> bodies = new Dictionary<String, Body>();
 
+                Logger logger = new Logger();
+
                 // Load all of the bodies
                 foreach (ConfigNode bodyNode in node.GetNodes(bodyNodeName))
                 {
@@ -221,9 +223,9 @@ namespace Kopernicus
 
                     try
                     {
-                        // Create a logger for this body
-                        Logger bodyLogger = new Logger(bodyNode.GetValue("name") + ".Body");
-                        bodyLogger.SetAsActive();
+                        // Create a logfile for this body
+                        logger.SetFilename(bodyNode.GetValue("name") + ".Body");
+                        logger.SetAsActive();
 
                         // Attempt to create the body
                         currentBody = new Body();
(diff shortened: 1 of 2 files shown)


src/Kopernicus/Configuration/Loader.cs
```diff
@@ -200,66 +200,91 @@ void IParserEventSubscriber.Apply(ConfigNode node)
...
+
+                        // Restore default logger
+                        bodyLogger.Flush();
+                        Logger.Default.SetAsActive();
                     }
                     catch (Exception e)
                     {
-                        bodyLogger.LogException(e);
-                        Logger.Default.Log("[Kopernicus]: Configuration.Loader: Failed to load Body: " + bodyNode.GetValue("name"));
+                        //bodyLogger.LogException(e);  -- This would be nice but becomes a problem when duplicate loggers are created: can't log then
```
StollD: Since you already check for a duplicated body before the try / catch block starts, this could stay as it is right now. The logger shouldn't throw because of duplicated names anymore, so it doesn't need to be created in a try / catch.

Corfiot:
The whole loading process seems to be aborted when a single log file fails to be created. This is why I covered it.

StollD:
Then please declare the bodyLogger variable outside of the try / catch but assign it inside. That way, the catch block can do a null-check on it and still log errors that are related to loading the body in the body log.

Corfiot:
The problem happened when the object was constructed not when it was set as active logger. Maybe we can create a simple constructor for it, then we can do as you suggest safely.

StollD:
Or we just do this

```
Logger bodyLogger;
try
{
    bodyLogger = new Logger("..");
    ...
}
catch (Exception e)
{
    if (bodyLogger != null) 
    {
        bodyLogger.LogException(e);
    }
    else
    {
        Logger.Default.LogException(e);
    }
    ...
}
```

Corfiot:
I'm not that good at C# yet, does a constructor that throw an exception set the instance to null?
Give me some time, I'll propose a new set that is similar and has the advantage or using a single Logger instance for all bodies, preventing possible garbage collection issues. I think that incorporates your strategy.

StollD:
> I'm not that good at C# yet, does a constructor that throw an exception set the instance to null?

Yes, thats what happens. The class cannot be initialized so the variable retains the previous value (which is null by default)

Corfiot:
Works exactly as before with less garbage collection, no null-checks and a smarter logger.

StollD:
Looks good, thank you!

Merging

Corfiot:
Thanks for the feedback.

PR #307, opened 2018-08-25
Added a way to change the light color in the main menu

Tutur716:


src/Kopernicus/Templates.cs
```diff
@@ -52,6 +52,9 @@ public class Templates
         // The body that should appear in MainMenu
         public static String menuBody { get; set; }
 
+        // The main menu color
+        public static ColorParser mainMenuLightColor = new ColorParser(new Color(1, 1, 1, 1));
```
StollD: You should just use a Color here, not a color parser.

Also, it would be awesome if the color in the main menu could be automatically fetched from the star the main menu body is orbiting. I am not sure if Kopernicus already has an API for that (probably `KopernicusStar.GetNearest(MainMenuBody).lightShifter.lightColor` or so), if not I can write down the exact code later.

src/Kopernicus/RuntimeUtility/RuntimeUtility.cs
```diff
@@ -424,6 +424,20 @@ void LateUpdate()
                     detail.stringValues = _details ?? (_details = Templates.PresetDisplayNames.ToArray());
                 }
             }
+
+            // Change main menu light color
+            if (HighLogic.LoadedScene == GameScenes.MAINMENU)
+            {
+                // I have to check two times if the game object exists, if i don't a nullref appears, and no-one likes nullref.
+                if(GameObject.Find("Directional light"))
```
StollD: You should probably cache that GameObject.

Tutur716:
Maybe it can be possible to do that with searching the first Star body main menu body's hierarchy..

StollD:
Yeah, thats what I tried to describe. `KopernicusStar.GetNearest()` basically does this, but it just checks for the distance between the two objects.

Poodmund:
I know it's not entirely helpful but Contract Configurator has a function in place where you can call the parent body, that body's star, then any children of that star etc. so it must have a similar API in place that would be required here.

Tutur716:
I don't really think that's really useful, sind since it's already very easy to loop parents and check of

src/Kopernicus/RuntimeUtility/RuntimeUtility.cs
```diff
@@ -357,11 +357,11 @@ void LateUpdate()
                 {
                     FieldInfo mode_f = typeof(OrbitTargeter).GetFields(BindingFlags.NonPublic | BindingFlags.Instance).FirstOrDefault(f => f.FieldType.IsEnum && f.FieldType.IsNested);
                     FieldInfo context_f = typeof(OrbitTargeter).GetFields(BindingFlags.NonPublic | BindingFlags.Instance).FirstOrDefault(f => f.FieldType == typeof(MapContextMenu));
-                    #if !KSP131
+#if !KSP131
```
StollD: revert those formatting changes please

Tutur716: I thought I didn't apply them... Alright

src/Kopernicus/RuntimeUtility/RuntimeUtility.cs
```diff
@@ -424,8 +424,69 @@ void LateUpdate()
...
+                                        }
+                                    }
+                                    if (mainCB)
+                                    {
+                                        Templates.mainMenuLightColor = KopernicusStar.GetNearest(mainCB).light.color;
+                                    }
+
+                                    else
+                                    {
+                                        Logger.Default.Log("Unable to find main menu body's celestialbody componant!");
```
StollD: Since this is executed at runtime you should probably use `Logger.Active.Log` to it ends up in `Kopernicus.Runtime.log` (created at RuntimeUtility.Awake() iirc)

Sigma88:
SQUAD removed the old Light object from the MainMenu so I doubt this solution would still work

however I think it's fair to say this is a feature we should look into adding to Kopernicus as a baseline, probably even automate the process to use the light from the parent star to the body displayed in the menu

Issue #309, opened 2018-09-02
New Orbit Icon customization tanks FPS in Planetarium Camera


Poodmund:
The new custom Orbit Icon feature introduced in 1.4.5-3 seems to tank the FPS when the Planetarium camera is active. 

Reproduction steps:
- Vanilla KSP
- Go to Tracking Station
- View FPS
---
- Vanilla KSP install + Kopernicus 1.4.5-4 + Dependencies
- Go to Tracking Station
- View FPS

See the difference. These are the values I pulled from my tests:
```
KSP 1.4.5 Stock                                         165fps (monitor limit)
KSP 1.4.5 + Kopernicus 1.4.5-4                          80fps
KSP 1.4.5 + Kopernicus 1.4.5-3                          85fps
KSP 1.4.5 + Kopernicus 1.4.5-2                          165fps (monitor limit)
KSP 1.4.5 + Kopernicus 1.4.5-4 w/ Orbit Icons Removed   165fps (monitor limit)
```
It seems that the custom Orbit Icon script added to Kopernicus.RuntimeUtility.cs is the cause as when it is removed and recompiled, the issue is not present. The code in question:
```
// Apply orbit icon customization
foreach (MapNode node in Resources.FindObjectsOfTypeAll<MapNode>())
{
    if (node.mapObject != null && node.mapObject.celestialBody != null && node.mapObject.celestialBody.Has("iconTexture"))
    {
        Texture2D texture = node.mapObject.celestialBody.Get<Texture2D>("iconTexture");
        node.SetIcon(Sprite.Create(texture,
            new Rect(0, 0, texture.width, texture.height),
            new Vector2(0.5f, 0.5f), 100, 1, SpriteMeshType.Tight,
            Vector4.zero));
    }
}
```
Referenced from this post in the Kopernicus thread: https://forum.kerbalspaceprogram.com/index.php?/topic/140580-145-4-131-13-kopernicus-kittopiatech/&do=findComment&comment=3443919

marr75:
[...]

PR #310, opened 2018-09-06
WIP: Orbital icon customization loop performance issues

marr75:
See #309 
By looping through just `FlightGlobals.Bodies` instead of all resources, this loop is quite a bit faster (doubled my framerate with GPP + GEP + OPM). I don't know if important bodies are excluded or if this can be optimized more aggressively without better test data (a config with customized icons, preferably a planet pack). Off the top of my head, a more aggressive optimization would be to identify a smaller set of customized bodies once and then loop through only those bodies OR determine if the icons can be set less frequently (only once, for example).

Very open to ideas and discussion. I'm not much of an artist/designer (or great with working with those assets) so I question how realistic an orbital icon test case I can put together without some outside guidance.

### DIFF
--- a/src/Kopernicus/RuntimeUtility/RuntimeUtility.cs
+++ b/src/Kopernicus/RuntimeUtility/RuntimeUtility.cs
@@ -51,6 +51,7 @@ namespace Kopernicus
     {
         // Variables
         public MapObject previous;
+        private readonly Dictionary<CelestialBody, Sprite> _spriteCache = new Dictionary<CelestialBody, Sprite>();
 
         // Awake() - flag this class as don't destroy on load and register delegates
         void Awake()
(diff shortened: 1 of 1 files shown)


marr75:
WIP update: appears to me this could be called once per body. Caveat: the method `UpdateType` is called every `LateUpdate` tick and this method might reset the icon back to an icon dependent upon the node's type if it detects a change in the icon type. Coincidentally, `SetIcon` calls `UpdateType` but does so *before* the icon is set to the sprite icon. I believe this would actually initialize the `MapNode` such that it `UpdateType` would effectively become a no-op.

StollD:
Looks good! Thanks for this, and sorry for me not looking at it for so long.

Issue #311, opened 2018-09-30
Palette .dds Loader doesn't function correctly


Poodmund:
I have been checking upon the functionality of the DDS 4bpp and 8bpp Palette texture format loading, that was incorporated previously, and it seems that it does not seem to work as intended.

I am unsure as to whether the files get loaded at all as when you set a .dds texture in either of these two formats as a colour map the body ends up solid white as shown in the image linked. If you set one as a Biome Map and then toggle the Biome Map overlay, the image does not render at all and the body scaled space texture is just shown.

[Image](https://media.discordapp.net/attachments/365954552354045953/495208988816834570/Untitled-1.png)

It seems that the loader is not loading the texture properly as can be seen in this part of the log:

```
[LOG 13:19:36.216] [OD] Adding for body Hale map named Hale_biome (CBRGB) for Hale of path = OPM/OPM_Textures/PluginData/Hale_biome.dds
[LOG 13:19:36.220] [Kopernicus]: Only DXT1, DXT5, A8, RGB24, RGBA32, RGB565, ARGB4444, RGBA4444, 4bpp palette and 8bpp palette are supported
[LOG 13:19:36.220] [Kopernicus]: failed to load F:/SteamLibrary/SteamApps/common/Kerbal Space Program/KSP_x64_Data/../GameData/OPM/OPM_Textures/PluginData/Hale_color.dds
[WRN 13:19:36.224] DontDestroyOnLoad only work for root GameObjects or components on root GameObjects.
[LOG 13:19:36.282] [OD] Adding for body Hale map named Hale_color (RGB) for Hale of path = OPM/OPM_Textures/PluginData/Hale_color.dds
```

[KSP Log](https://cdn.discordapp.com/attachments/365954552354045953/495209148057780224/KSP.log)
[Output Log](https://cdn.discordapp.com/attachments/365954552354045953/495209377234419741/output_log.txt)
[MM Cache](https://cdn.discordapp.com/attachments/365954552354045953/495209568381435924/ModuleManager.ConfigCache)
[Kopernicus Body Log](https://cdn.discordapp.com/attachments/365954552354045953/495209569023295489/Hale.Body.log)

If you need a texture in this format to test with, just shout and I'll link one.

Sigma88:
This has been fixed, should work fine from the next Kopernicus update

a small note:

dds images saved with mipmaps will be compatible with this system, however the mipmaps will not be used. this means that mipmaps are a waste of memory space

[...]

PR #312, opened 2018-10-03
fix dds palette loader

Sigma88:


Sigma88:
this should solve #311

PR #313, opened 2018-10-03
add support for mipmaps (kind of)

Sigma88:


Sigma88:
this should add support for images using dds palette with mipmaps

the mipmaps are not used, the code just ignores them now, instead of giving up too early

PR #328, opened 2018-10-26
Use Vector3d instead of Vector3

sswelm:
by using vector3d instead of Vector3, the calculated distance is much more accurate, resulting in much less instability in the final solar power

### DIFF
--- a/src/Kopernicus.Components/KopernicusStar.cs
+++ b/src/Kopernicus.Components/KopernicusStar.cs
@@ -155,10 +155,10 @@ namespace Kopernicus
                 // Get sunVector
                 RaycastHit raycastHit;
                 Boolean directSunlight = false;
-                Vector3d scaledSpace = ScaledSpace.LocalToScaledSpace(fi.IntegratorTransform.position);
+                Vector3d scaledSpace = ScaledSpace.LocalToScaledSpace(fi.Vessel.CoMD);
                 Double scale = Math.Max((star.sun.scaledBody.transform.position - scaledSpace).magnitude, 1);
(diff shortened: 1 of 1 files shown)


src/Kopernicus.Components/KopernicusStar.cs
```diff
@@ -155,10 +155,10 @@ public static Double Flux(ModularFlightIntegrator fi, KopernicusStar star)
                 // Get sunVector
                 RaycastHit raycastHit;
                 Boolean directSunlight = false;
-                Vector3d scaledSpace = ScaledSpace.LocalToScaledSpace(fi.IntegratorTransform.position);
+                Vector3d scaledSpace = ScaledSpace.LocalToScaledSpace(fi.Vessel.CoMD);
```
StollD: Is CoMD actually the same as the vessel position in world space? I am not entirely sure, and can't check it at the moment.

src/Kopernicus.Components/KopernicusStar.cs
```diff
@@ -155,10 +155,10 @@ public static Double Flux(ModularFlightIntegrator fi, KopernicusStar star)
...
                 Double scale = Math.Max((star.sun.scaledBody.transform.position - scaledSpace).magnitude, 1);
-                Vector3 sunVector = (star.sun.scaledBody.transform.position - scaledSpace) / scale;
-                Ray ray = new Ray(ScaledSpace.LocalToScaledSpace(fi.IntegratorTransform.position), sunVector);
+                Vector3d sunVector = (star.sun.position - scaledSpace) / scale;
```
StollD: This change would have serious implications on the calculation, because you are replacing a scaled space position with a local space one (scaled space = local space / 6000).

PR #336, opened 2018-12-17
fix editor ground

Sigma88:
I left this as a pull request since it is a pretty substantial change and I'm not sure you agree it is needed

these changes fix the recoloring of ground in the editor scenes

it will work automatically without the need of user input, but it allows for a certain degree of customization that was lacking before.


Sigma88:
I was playing around with Kerbal Konstructs and noticed that this method fails if KK is installed

@GER-Space does KK change the name of some materials used in the space center?

GER-Space:
KK should not alter the ksc buildings, but any instances placed of the buildings can have new gras textures. 
Is there a easy way to recreate that error, so I can debug this? 
Sorry I'm a kopernicus noop.

Sigma88:
I will take a closer look and let you know what really happens 

for now I don't think any actions are needed since we still have a way to make it work

StollD:
Is this ready to be merged? Or do you still want to make changes / test things?

Sigma88:
I think this is ready to be merged

PR #353, opened 2019-04-28
Update HazardousOcean.cs

TheWhiteGuardian:
Currently HazardousOcean adds temperature per _frame_ and not per _second._ Not only is degrees per second a much easier unit to work with, it will also remain consistent with respect to lag spikes. Besides that, it'll make even the tiniest of HazardousOcean temperature emissions impossible to deal with for beefy computers, as even the slightest additions will hit like a truck when multiplied by 120 as a consequence of running at 120 FPS.

### DIFF
--- a/src/Kopernicus.Components/HazardousOcean.cs
+++ b/src/Kopernicus.Components/HazardousOcean.cs
@@ -71,7 +71,7 @@ namespace Kopernicus
                     Double distanceToPlanet = Math.Abs(Vector3d.Distance(vessel.transform.position, body.transform.position)) - ocean.GetSurfaceHeight(ocean.GetRelativePosition(vessel.transform.position));
                     Double heatingRate = heatCurve.Evaluate((Single)distanceToPlanet);
                     foreach (Part part in vessel.Parts)
-                        part.temperature += heatingRate;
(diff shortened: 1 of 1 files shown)


DeltaDizzy:
HazardousOcean no longer exists. It has been replaced with HazardousBody. HB allows you to control the heat at specific altitudes, latitudes, and longitudes. It also contains an interval value, which controls the time between heat applications, but I am not sure what units it uses.

TheWhiteGuardian:
Literally just noticed the 'obsolete' notice.
Allow me to move this pull request to r/facepalm.

StollD:
Am Sonntag, 28. April 2019, 21:06:57 CEST schrieb TheWhiteGuardian:
> Literally just noticed the 'obsolete' notice.
> Allow me to move this pull request to r/facepalm.

I can imagine some mods still using HazardousOcean, so I 
don't see any problem in merging this.

If you have the time and motivation I would appreciate if you could make sure 
though that the FPS problems you mentioned with HazardousOcean don't happen 
with HazardousBody and an otherwise identical config.

TheWhiteGuardian:
As requested, I wrote a quick fix that solves the FPS for both the new HazardousBody and old HazardousOcean scripts. Already submitted a new pull request from a fork based off the current master branch.

PR #354, opened 2019-04-28
Fix FPS for Hazardous-components

TheWhiteGuardian:
Temperature-per-second fix for both HazardousBody and the legacy HazardousOcean components.

### DIFF
--- a/src/Kopernicus.Components/HazardousBody.cs
+++ b/src/Kopernicus.Components/HazardousBody.cs
@@ -67,6 +67,13 @@ namespace Kopernicus
             public MapSO HeatMap;
 
             private CelestialBody _body;
+			
+			// How far is the script in the HeatInterval cycle?
+			private Single IntervalPosition = 0f;
+			
+			// precompute the division: typically you can run 100 multiplications in the time it takes for one division operator to finish.
+			// we might as well allocate 32 more bits per planet for the sake of performance.
+			private Single Frequency;
             
             /// <summary>
             /// Get the body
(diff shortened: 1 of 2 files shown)


StollD:
Sorry to be that guy, but is `Time.deltaTime` really neccessary for 
HazardousBody? If I remember correctly, it was used to smooth out operations 
that run every frame so they effectively take one second to complete, but since 
HazardousBody runs in a Coroutine it just tells Unity to wat one (or n, 
depending on the configuration) seconds.

It would probably work if you set the second value to one, but I fear it might 
break if it is set to anything <> 1 (like the default config which I think just 
runs 20 times per second.)

Definitly appreciate the efford though, thank you.

TheWhiteGuardian:
Uhp, missed the fact that you tell the Coroutine to yield return 'WaitForSeconds' to specify a flexible update interval. No, in that case Time.deltaTime isn't needed. Instead heat should be divided by the interval duration.

Though I wonder why you've chosen to have every instance of HazardousBody maintain a coroutine. Coroutines have a significant memory overhead and take some effort to start and stop.

I'm going to see if I can devise a more flexible system using a manager class to handle the counting and using a UnityEvent to handle all update cycles.

I wonder though, does KSP actually destroy parts for overheating if they're not on the active vessel? 

Anyhow, can I still edit the changes I'm proposing in the pull request after making said pull request?

StollD:
Am Sonntag, 28. April 2019, 21:44:33 CEST schrieb TheWhiteGuardian:
> Though I wonder why you've chosen to have every instance of HazardousBody
> maintain a coroutine. Coroutines have a significant memory overhead and
> take some effort to start and stop.
Probably because the code is easier to write, and the overhead is minimal, 
compared to how often it gets used in my experience.

> I wonder though, does KSP actually destroy parts for overheating if they're
> not on the active vessel?
All vessels that are not within the loaded physics range (2km or so) get 
unloaded and only have their on-rails orbits simulated

> Anyhow, can I still edit the changes I'm proposing in the pull request after
> making said pull request?
Just push more changes to the branch you made the pull request from. It's 
really just "take everything that is in here and not in master and ask to 
merge it in".

TheWhiteGuardian:
Gave HazardousBody a rewrite to omit the coroutine, because using 32 more bits is easier for Unity especially if we use it multiple times per frame. It should also scale heat correctly now.

TheWhiteGuardian:
Made an updated version that keeps the coroutine intact. I'm still concerned about using Coroutines though, mainly because of this article by Unity:
https://unity3d.com/learn/tutorials/topics/performance-optimization/optimizing-garbage-collection-unity-games
Specifically the header 'Coroutines'.

StollD:
Unfortunately I completely messed up my local git history while doing code refactoring, up to a point where I had to force push the repository to not risk loosing my gargantuan commit. This means I had to delete the commits from this PR.

Sadly I cannot remerge them either because after my gigantic commit everything is a bit different, and I would effectively be me rewriting them (and showing up as the commiter).

If you have the time and motivation to rebase those patches, I would be happy to re-merge them, otherwise I will just readd them myself. Sorry again for messing that up.

PR #364, opened 2019-06-11
Avoid replacing ModuleDeployableSolarPanel

Sigma88:


Sigma88:
updated to avoid patching all PART nodes

PR #368, opened 2019-07-10
Fix "selectable"

Sigma88:
should be safe to merge

### DIFF
--- a/src/Kopernicus/Configuration/Body.cs
+++ b/src/Kopernicus/Configuration/Body.cs
@@ -94,7 +94,10 @@ namespace Kopernicus.Configuration
                 CelestialBody.Set("barycenter", value.Value);
 
                 // Reuse the selectable option, to reduce the amount of boilerplate code we need.
-                CelestialBody.Set("selectable", !value.Value);
+                if (value.Value)
+                {
+                    CelestialBody.Set("selectable", !value.Value);
+                }
             }
         }
 
--- a/src/Kopernicus/Configuration/Body.cs
+++ b/src/Kopernicus/Configuration/Body.cs
@@ -94,7 +94,7 @@ namespace Kopernicus.Configuration
                 CelestialBody.Set("barycenter", value.Value);
 
                 // Reuse the selectable option, to reduce the amount of boilerplate code we need.
-                CelestialBody.Set("selectable", value.Value);
+                CelestialBody.Set("selectable", !value.Value);
             }
         }
 
--- a/src/Kopernicus/RuntimeUtility/RuntimeUtility.cs
+++ b/src/Kopernicus/RuntimeUtility/RuntimeUtility.cs
@@ -399,8 +399,7 @@ namespace Kopernicus.RuntimeUtility
             {
                 trackingstation.Add(PlanetariumCamera.fetch.targets.Find(t => t.celestialBody == cb));
             });
-            PlanetariumCamera.fetch.targets = trackingstation.Where(m =>
-                m?.celestialBody != null && m.celestialBody.Get("selectable", true)).ToList();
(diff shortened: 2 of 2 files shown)


Sigma88:
the issue was that
```
Utility.DoRecursive(PSystemManager.Instance.localBodies[0], cb => cb.orbitingBodies, cb =>
{
    trackingstation.Add(PlanetariumCamera.fetch.targets.Find(t => t.celestialBody == cb));
});
```
was adding null objects to the list, by checking for `m?.celestialBody != null` this is fixed

Sigma88:
to be hones, using
```
Where(m => m?.celestialBody != null && m.celestialBody.Get("selectable", true))
```
is redundant and it should be enough to use
```
Where(m => m != null)
```

let me know if you prefere the latter

Sigma88:
I updated the code to be more streamlined,

I would merge this directly but I had another doubt about the Body parser:

currently the [barycenter parser](https://github.com/Kopernicus/Kopernicus/blob/selectable-fix/src/Kopernicus/Configuration/Body.cs#L87-L99) looks like this:

```
[ParserTarget("barycenter")]
[KittopiaHideOption]
public NumericParser<Boolean> Barycenter
{
	get { return CelestialBody.Get("barycenter", false); }
	set
	{
		CelestialBody.Set("barycenter", value.Value);

		// Reuse the selectable option, to reduce the amount of boilerplate code we need.
		CelestialBody.Set("selectable", !value.Value);
	}
}
```

I don't like the fact that the `set` will add `selectable = true` to all bodies that have `barycenter = false`, and to a lesser extent I don't like the fact that `barycenter = false` itself will be added.

so, I propose to go with either of these changes:

A:
```
set
{
	CelestialBody.Set("barycenter", value.Value);

	// Reuse the selectable option, to reduce the amount of boilerplate code we need.
	if (value.Value)
	{
		CelestialBody.Set("selectable", !value.Value);
	}
}
```

or B:

```
set
{
	if (value.Value)
	{
		CelestialBody.Set("barycenter", value.Value);

		// Reuse the selectable option, to reduce the amount of boilerplate code we need.
		CelestialBody.Set("selectable", !value.Value);
	}
}
```

Sigma88:
if you like A, there is also this option in case you like the indentation better:

```
set
{
	CelestialBody.Set("barycenter", value.Value);

	// Reuse the selectable option, to reduce the amount of boilerplate code we need.
	if (value.Value)
		CelestialBody.Set("selectable", !value.Value);
}
```

or 

```
set
{
	CelestialBody.Set("barycenter", value.Value);

	// Reuse the selectable option, to reduce the amount of boilerplate code we need.
	if (value.Value) CelestialBody.Set("selectable", !value.Value);
}
```

StollD:
> I don't like the fact that the `set` will add `selectable = true` to
> all bodies that have `barycenter = false`, and to a lesser extent I
> don't like the fact that `barycenter = false` itself will be added.

Well, tbh, since `barycenter = False` is the default, it should never 
actually get added to a config (unless someone wants to change a 
barycenter into a real body with MM), I don't think that is much of a 
problem. And `barycenter` should not be used anymore anyways.

But if you *want* to change it, feel free. I am ok with all of the 
variants you posted.

Thank you again for looking into this and fixing it.

Sigma88:
>But if you *want* to change it, feel free. I am ok with all of the
variants you posted.

it's not that I \*want\* to change it, but it seems to me that it would be safer to at least avoid setting selectable to true

this because if some other plugin (or another part of kopernicus in the future) is going to set `selectable = false` at an earlier time, I wouldn't want the parser for `barycenter` to overwrite that setting

I can see an argument for actually storing `barycenter = false`, because if someone goes out of his way to add "barycenter = false" he would expect that to be saved onto the body.

PR #378, opened 2019-08-26
Update SolarPanels.cfg

HSJasperism:
Reduce the inflated number of parts allegedly patched by Module Manager (make it so that this patch only goes through parts with specific modules). The change in syntax requires later versions of MM but Kopernicus already requires that. Assumption is that parts can only have one ModuleDeployableSolarPanel

### DIFF
--- a/build/GameData/Kopernicus/Config/SolarPanels.cfg
+++ b/build/GameData/Kopernicus/Config/SolarPanels.cfg
@@ -7,7 +7,7 @@
     // If you want to keep your ModuleDeployableSolarPanel, add "useKopernicusSolarPanels = false" to the MODULE node
     // That will stop Kopernicus from replacing it
     //
-    @MODULE[ModuleDeployableSolarPanel]:HAS[#useKopernicusSolarPanels[*]]
+    @MODULE:HAS[#useKopernicusSolarPanels[*]]
     {
         @useKopernicusSolarPanels,* ^= :F:f:
         @useKopernicusSolarPanels,* ^= :A:a:
@@ -15,7 +15,7 @@
         @useKopernicusSolarPanels,* ^= :S:s:
         @useKopernicusSolarPanels,* ^= :E:e:
     }
-    @MODULE[ModuleDeployableSolarPanel]:HAS[~useKopernicusSolarPanels[false]]
+    @MODULE:HAS[#name[ModuleDeployableSolarPanel],~useKopernicusSolarPanels[false]]
     {
         @name = KopernicusSolarPanel
     }
--- a/build/GameData/Kopernicus/Config/SolarPanels.cfg
+++ b/build/GameData/Kopernicus/Config/SolarPanels.cfg
@@ -1,11 +1,13 @@
-@PART:FINAL
+@PART[*]:HAS[@MODULE[ModuleDeployableSolarPanel]]:FINAL
 {
+    //
     // This will replace all instances of ModuleDeployableSolarPanel with the Kopernicus version
     // that has proper support for multiple lightsources
     // 
     // If you want to keep your ModuleDeployableSolarPanel, add "useKopernicusSolarPanels = false" to the MODULE node
     // That will stop Kopernicus from replacing it
-    @MODULE:HAS[#useKopernicusSolarPanels[*]]
+    //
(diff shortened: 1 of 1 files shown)


Sigma88:
hmm I thought I had already did that 

weird

I looked at the changes and they look fine

I will leave that to @StollD  to decide if he wants to merge this or my other PR

HSJasperism:
> 
> 
> hmm I thought I had already did that
> 
> weird
> 
> I looked at the changes and they look fine
> 
> I will leave that to @StollD to decide if he wants to merge this or my other PR

Hey, I didn't notice https://github.com/Kopernicus/Kopernicus/pull/364 on mobile.
The changes I suggested for the old Line 16 should be replaced with your version.
The changes to Line 1 just aim to make sure Kopernicus skips parts without the relevant module - so they don't show up on the patches applied count on the loading screen as well as remove mentions of Kopernicus modifying non-solar panel parts from the log file. 
Your changes are better long term, and helps with people enabling and disabling the Solar Panel feature.

Sigma88:
that's what I meant, I thought I had already included your changes to line 1 into kopernicus however for some reason they are not there.

I do agree that, regardless if my proposed changes in #364 are accepted or not, your changes to line 1 should be included.

@StollD  I would suggest accepting this PR if you are still not sure about my PR

build/GameData/Kopernicus/Config/SolarPanels.cfg
```diff
@@ -1,19 +1,21 @@
...
+    //
+    @MODULE[ModuleDeployableSolarPanel]:HAS[#useKopernicusSolarPanels[*]]
     {
         @useKopernicusSolarPanels,* ^= :F:f:
         @useKopernicusSolarPanels,* ^= :A:a:
         @useKopernicusSolarPanels,* ^= :L:l:
         @useKopernicusSolarPanels,* ^= :S:s:
         @useKopernicusSolarPanels,* ^= :E:e:
     }
-    @MODULE:HAS[#name[ModuleDeployableSolarPanel],~useKopernicusSolarPanels[false]]
```
Sigma88: I would rather keep `:HAS[#name[]]` rather than just using `@MODULE[]`

Sigma88:
there was also a suggestion to change `:FINAL` into `:LAST[Kopernicus]`

however I never heard @StollD 's opinion on that.


this PR looks fine to me and can be merged if #364  is rejected

build/GameData/Kopernicus/Config/SolarPanels.cfg
```diff
@@ -1,10 +1,12 @@
-@PART:FINAL
+@PART[*]:HAS[@MODULE[ModuleDeployableSolarPanel]]:FINAL
```
Sigma88: I would also remove the `[*]` after `PART`

I don't think it adds anything to have it there except for (maybe?) triggering an unnecessary check

Sigma88: also, I don't know the difference between

```
@Kopernicus:HAS[@MODULE[ModuleDeployableSolarPanel]]:FINAL
```
and
```
@Kopernicus:HAS[@MODULE:HAS[#name[ModuleDeployableSolarPanel]]]:FINAL
```

I usually go for the latter, but that's only because it avoided some very old MM bugs that have probably been patched by now. However it has been a long time since I tested it so 🤷‍♂

DeltaDizzy:
I realize I am very late to this, but is the assumption believed to be true or simply used due to the alternative adding too much complexity? The reason I bring this up is that not only can parts have multiple instances of MDSP, but several (modded) ones do. It is most often done for parts that have curved/cylindrical solar panels in mods that don't want to make Near Future Solar Core a dependency. [An example is the Pioneer 6 from Bluedog Design Bureau.](https://github.com/CobaltWolf/Bluedog-Design-Bureau/blob/77316f62b126735047e38ca83cad3fd5df243a55/Gamedata/Bluedog_DB/Parts/ProbeExpansion/bluedog_Pioneer6_Service.cfg#L52-L116)

Sigma88:
I didn't know this could be the case, I usually try to account for this kind of things but in this case I don't think the code will work properly. However I think I can make it work.

Sigma88:
I will close this PR since the changes required by @HSJasperism  have been included in the new code.

@DeltaDizzy  you can take a look at #393  for what you were asking before

Issue #386, opened 2019-10-12
Occult Octaves Occurences


Sigma88:
I have been investigating an issue that a lot of users were reporting on installs with SD + AKR

This sent me on a treasure hunt that ultimately led me to believe it could be caused by the fact that some PQSMod loaders are not loading octaves "properly"

safe PQSMods seem to use this code:

```
        [ParserTarget("octaves")]
        public NumericParser<Int32> Octaves
        {
            get { return Mod.octaves; }
            set { Mod.octaves = Mathf.Clamp(value, 1, 30); }
        }
```

note: I only tested [VertexHeightNoise](https://github.com/Kopernicus/Kopernicus/blob/master/src/Kopernicus/Configuration/ModLoader/VertexHeightNoise.cs#L58-L63) but I think it's safe to assume all mods with this code will not generate any errors

<br><br><br>

the bug was caused by [VertexSimplexHeight](https://github.com/Kopernicus/Kopernicus/blob/master/src/Kopernicus/Configuration/ModLoader/VertexSimplexHeight.cs#L55-L60) that has this code:

```
        [ParserTarget("octaves")]
        public NumericParser<Double> Octaves
        {
            get { return Mod.octaves; }
            set { Mod.octaves = value; }
        }
```

as you can see this is using double instead of Int and it doesn't clamp the value

PR #387, opened 2019-10-12
Proposed fix for #386

Sigma88:
this might be the solution to issue #386

### DIFF
--- a/src/Kopernicus/Configuration/ModLoader/VertexPlanet.cs
+++ b/src/Kopernicus/Configuration/ModLoader/VertexPlanet.cs
@@ -67,10 +67,10 @@ namespace Kopernicus.Configuration.ModLoader
 
             // octaves
             [ParserTarget("octaves")]
-            public NumericParser<Int32> Octaves
+            public NumericParser<Double> Octaves
             {
                 get { return Value.octaves; }
-                set { Value.octaves = Mathf.Clamp(value, 1, 30); }
+                set { Value.octaves = value > 30 ? 30 : value > 1 ? value : 1; }
             }
 
             // persistance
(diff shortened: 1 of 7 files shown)


Sigma88:
I had to change the code to use

```value > 30 ? 30 : value > 1 ? value : 1;```

because apparently KSP mods still have some of those values as Double

valid alternatives could be:


```
        [ParserTarget("octaves")]
        public NumericParser<Int32> Octaves
        {
            get { return (Int32)Mod.octaves; }
            set { Mod.octaves = Mathf.Clamp(value, 1, 30); }
        }
```

or adding an extension for clamping doubles like this:

```
        public static Double Clamp(this NumericParser<Double> value, Double min, Double max)
        {
            if (value > max)
                return max;

            if (value > min)
                return value;

            return min;
        }
```

which would then result in the code for the octaves to look like this:

```
        [ParserTarget("octaves")]
        public NumericParser<Double> Octaves
        {
            get { return Mod.octaves; }
            set { Mod.octaves = value.Clamp(1, 30); }
        }
```

StollD:
I haven't checked the KSP API yet, but if the datatypes of the parsers don't match the ones of the values behind them, the parsers should be adjusted. If they are doubles, we should just use `Math.Clamp` instead of `Mathf.Clamp`.

Sigma88:
ok

I will update the code, however I don't think that it will be enough

Sigma88:
I made the changes, however I think this will still result in an error if the "octaves" parameter is missing in the cfg

Sigma88:
nice!

I'll delete the branch because we already have too many of them 🤣

PR #388, opened 2019-10-13
Avoid replacing ModuleDeployableSolarPanel

Sigma88:
this will introduce a new approach to managing kopernicus solarpanels

### DIFF
--- a/build/GameData/Kopernicus/Config/SolarPanels.cfg
+++ b/build/GameData/Kopernicus/Config/SolarPanels.cfg
@@ -1,4 +1,4 @@
-@PART:FINAL
+@PART:HAS[@MODULE:HAS[#name[ModuleDeployableSolarPanel]]]:FINAL
 {
     // This will replace all instances of ModuleDeployableSolarPanel with the Kopernicus version
     // that has proper support for multiple lightsources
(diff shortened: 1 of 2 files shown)


StollD:
I probably already asked you about this at some point but I forgot: Have you tested if this breaks solar panels the same way that installing Kopernicus into a stock game breaks them?

Also, sorry for totally forgetting about these changes for so long

Sigma88:
I think this would be safer, I had tested it at some point and found that even launched crafts were updated to get the kopernicus solar panel

StollD:
Yes, but what about crafts who already have the current KopernicusSolarPanel, and who would be reverted to the stock one? I get that this is way better than the old way for *new* kopernicus installs. 

So basically:
* you start a savegame 100% stock and put a vessel with a solar panel into orbit
* You install the latest Kopernicus release -> the solar panel breaks
* You put up another vessel, this time it already has the Kopernicus solar panel
* Now you install the code from this PR, and revert the second vessel to the stock module. 

I am curious what happens with the second vessel now.

Sigma88:
I tested this a long time ago so take this with a grain of salt

I remember testing both issues and the results were:

- existing stock craft will not have issues once you install kopernicus

- existing kopernicus crafts will not have issues once you uninstall kopernicus

I might do another test just to make sure everything is fine and I'm not misremembering

if you have other tests you would like me to do just let me know

StollD:
> existing kopernicus crafts will not have issues once you uninstall kopernicus

Yes, but what I am curious about is whether existing Kopernicus crafts will have issues will have issues once you update from an earlier version to a version that contains this change.

I.e. if this will break existing saves if / when it gets released.

Sigma88:
I think I had tested that as well, I'll test it again just in case and I will let you know

StollD:
Merged manually into the 1.8 branch, thank you.

Issue #398, opened 2019-11-18
space center grass color doesn't change in 1.8.1


Sigma88:
I'll check that and try to fix it myself

(reported by @GregroxMun)

GER-Space:
The buildings got a new grass shader and textures. 
This requires some new color management and a new target variable for the color.

`_GrassColor ` instead of `_Color`

The way the color is applied changed from `lerp(tex, color.rgb, color.a)` to `tex.rgb*col.rgb`

Also it got a bunch of new map names for the textures.
Ping me if you need the names, I'm only on mobile and it's painful to type for now.

Sigma88:
yes I already have a fix ready, but thanks for the heads up :)

for references these are the options I am considering introducing into kopernicus

```
SpaceCenter
{
    groundColor = // Color (For backward compatibility, will change the "_GrassColor")
    groundTexture = // Texture (For backward compatibility, will change the "_NearGrassTexture")
    GrassMaterial
    {
        nearGrassTexture = // Texture
        nearGrassTextureOffset = // Vector2
        nearGrassTextureScale = // Vector2
        nearGrassTiling = // Float
        farGrassTexture = // Texture
        farGrassTextureOffset = // Vector2
        farGrassTextureScale = // Vector2
        farGrassTiling = // Float
        farGrassBlendDistance = // Float
        grassColor = // Color
        tarmacTexture = // Texture
        tarmacTextureOffset = // Vector2
        tarmacTextureScale = // Vector2
        tarmacColor = // Color
        opacity = // Float
        rimColor = // Float
        rimFalloff = // Float
        underwaterFogFactor = // Float
    }
}```

there are still some kinks that we need to figure out because it seems like the tarmac texture needs to have different scales depending on which part of the ksc is colored, so I need a neat way to figure out how to account for that

GER-Space:
this list looks complete. 

1. You could ignore tarmacTextureOffset and Tiling. No reason to mess with that. 
2. I made a list of the materials of the KSC inclusing the tarmac Tiling: 
https://github.com/GER-Space/Kerbal-Konstructs/blob/master/Docs/KSC%20Materials.xlsx
You could use a dictionary for the material-name to tiling or only expose this parameter as a multiplication factor of the stock tiling. (if you want this to be edible)
3. tarmacColor is available, but has no effect. If you want I can supply a replacement shader which implement the tarmacColor parameter, because I watented/needed that for Kerbal-Konstructs. 
[...]

PR #400, opened 2019-11-23
GrassMaterial

Sigma88:
This fixes the KSC grass color change both for the PQS city and inside the Editor

please @StollD  review this since it has a lot of big changes that I am not sure you are ok with


src/Kopernicus/Configuration/SpaceCenterLoader.cs
```diff
@@ -415,6 +419,7 @@ public Vector2Parser EditorGroundTexOffsetParser
         // Apply event
         void IParserEventSubscriber.Apply(ConfigNode node)
         {
+            KSC.Instance = Value;
```
StollD: This shouldn't be neccessary. `KSC.Instance` is set when the loading was completed (to whatever space center was found on the home body) and while loading all calls to it are guarded with `if (!Injector.IsPrefab)` checks.

EDIT: I see the problem, you are storing everything in the `KSC` object, and that needs to be passed through to the `GrassMaterialLoader`. I think this could be solved by creating the `GrassMaterialLoader` ourselves in the constructor, instead of letting the parser do it. Then the `KSC` object could be passed through the constructor.

I just don't like the parser modifying global states, unless it is absolutely neccessary.

src/Kopernicus/Configuration/SpaceCenterLoader.cs
```diff
@@ -366,6 +366,10 @@ public ColorParser EditorGroundColorParser
             }
         }
 
+        [ParserTarget("GrassMaterial", AllowMerge = true)]
```
StollD: I would prefer this to be just `Material` like it is in most if not all other cases in Kopernicus.

Sigma88: I wanted to make a distinction from "SpaceCenter/Material" since this material only applies to the grass. however, I can change it to Material if you prefere

src/Kopernicus/Configuration/SpaceCenterLoader.cs
Sigma88: I'll do that if you prefere

Sigma88:
I applied the changes you asked, I'm not sure I handled that properly tho... please give it a double check

Sigma88:
@gameslinx did some testing (thanks GL by the way) and he found out that different parts of the KSC have different scales for the tarmac texture

I am in the process of figuring out how I want to approach this, if you have any ideas feel free to let me know

Sigma88:
Some new commits:

https://github.com/Kopernicus/Kopernicus/pull/400/commits/10508fd4d3b4630b1eeb2f79d6db64e6496c6c21 To avoid nullrefs if the material node is missing from the cfg

https://github.com/Kopernicus/Kopernicus/pull/400/commits/7fa33e61723ae0263db88dfab66b2bdf7135f8de To make sure the scale difference required by the stock materials is conserved

according to @Ger-Space _TarmacColor has no effect (https://github.com/Kopernicus/Kopernicus/issues/398#issuecomment-564278093) so I will probably remove it after I confirm this

also, I'm not sure if I want to keep the "tiling" and "offset" options. Offset at least feels like it might be useful, but tiling I have no idea

src/Kopernicus/Components/KSC.cs
```diff
@@ -100,6 +100,10 @@ public class KSC : SerializableMonoBehaviour
         [SerializeField]
         public Color? color;
 
+        // Grass Material
+        [SerializeField]
+        public GrassMaterial Material = new GrassMaterial();
```
Kerbas-ad-astra: I've seen other commits to the effect that behavior like this causes exceptions in KSP 1.8's version of Unity (e.g. https://github.com/snjo/Firespitter/pull/219/files#diff-f12b9be8c1e03fbc2ba80828371d75f4).  So maybe this should have ```public GrassMaterial Material;``` up here and then instantiate ```material = new GrassMaterial();``` in ```Awake()```?

Sigma88: sure, seems reasonable

src/Kopernicus/Configuration/SpaceCenterLoader.cs
Sigma88: this has been changed

src/Kopernicus/Configuration/SpaceCenterLoader.cs
Sigma88: I think this is fixed now, I'll leave the PR open for you to review

src/Kopernicus/Components/KSC.cs
Sigma88: fixed

Sigma88:
this should be ready now

PR #401, opened 2019-11-23
Clean Logs

Sigma88:
when running kopernicus, old Logs remain in the Logs folder and are not
properly cleaned
this change makes sure every time the whole Logs folder is cleared


Sigma88:
I initially had changed the LogAggregator.cs file, but then I decided to move the clearing of the folder to the Logger.cs file instead.

StollD:
Yeah, if you would add it to LogAggregator, it would delete the Logs folder every time KSP is closed and then zip an empty folder.

PR #409, opened 2020-01-17
Hazardous body

Sigma88:


Tuupertunut:
Now I'm confused. This code applies heating in a totally different way, which I don't quite understand.

It seems to me that you are trying to get the temperature to rise slower at higher temperatures. The thing is: this already happens using constant heating. With this model you are actually overcompensating the slowing of the rise of temperature at high temperatures. That is, unless you are trying to achieve something completely different.

Sigma88:
the previous way of applying heat was disabled during timewarp, which was something I wanted to avoid.

Once I let the heat continue during timewarp it was clear that the old method would just heat the parts with no limit. 

to avoid this I changed the heating from:

```new temperature = old temperature + heat/interval```

to 

```new temperature = old temperature + ((max temperature - old temperature) * 0.69420 / lambda) * TimeWarp.fixedDeltaTime```

where "lambda" it's the time it takes for the temperature delta `(max temperature - part temperature)` to be reduced by half

this will result in this type of curve:

image

Tuupertunut:
You have discovered that ksp thermal physics is not running in timewarp. I don't know if that's entirely correct, but let's assume it is.

What you first did was that you added HazardousBody heating to timewarp. Normally in ksp there is heat added from the sun and atmospheres and it is dissipated through thermal conduction and radiation. Now because there is no stock ksp thermal physics in timewarp, you ran in the situation where HazardousBody was the only thermal system running and the heat was not dissipated by normal means. That's probably why HazardousBody heating was previously disabled in timewarp.

In my opinion, disabling heating in timewarp is correct, because it's consistent with the ksp stock thermal systems. If you want to introduce proper thermal physics in timewarp, you will have to replicate the entire ksp thermal module.

Now your solution was to create a curve that approximately models the temperature if there is a single heat source and constant rate of dissipation. In ksp however, it will also have to interact with other heating mechanisms the game provides. The solution has some major problems:
- In timewarp it doesn't account for any other heat sinks/sources than HazardousBody. For example, radiators won't be able to cool down the applied heat any more effectively than no radiators. The key here is that your curve models **constant dissipation** (radiators won't dissipate any faster) and **single heat source** (sun is not taken into account).
- In non-timewarp it applies heating incorrectly. It artificially caps all temperature to maxTemp, even if heated by the sun. It also dissipates heat too fast because there is now both stock ksp dissipation and the dissipation built-in to the curve.

Having a constant heating rate like before is a simple and physically correct way to achieve this and it fits well with the ksp stock systems. I fear that heating just has to be disabled for timewarp as there is no easy way to correctly handle it.

Sigma88:
> Having a constant heating rate like before is a simple and physically correct way to achieve this and it fits well with the ksp stock systems. I fear that heating just has to be disabled for timewarp as there is no easy way to correctly handle it.

following my tests this is what was happening:

1. with a constant temperature increase and no heating in timewarp (like it was before), as soon as you hit timewarp the temperature almost immediately dropped back to the temperature the environment has without HazardousBody

2. with a constant temperature increase and heating enabled in timewarp, the temperature keeps increasing as expected but it never stops.

3. with this new system, the temperature increases following the curve in my previous message


All the tests were done without any radiators, I will try using a stronger HazardousBody setting to check what happens when you rely on radiators to remain at a safe temperature and you activate timewarp

Tuupertunut:
Number 1. raises an interesting point: If the temperature drops back to environment temp but not some constant everywhere, it means that ksp is doing _some_ thermal physics on timewarp. It's still not doing _all_ of it as seen from heat dissipation not working.

Sigma88:
there seems to be another issue, part.temperature will increase mostly the internal temperature while the skin temperature will still get hotter but not as much as the internal temperature

this should be fixed because the heat is coming from outside the part

Tuupertunut:
If you want you can just release 1.8 without this patch and we will have more time to look at it.

Sigma88:
so, I realized Kopernicus already requires MFI so might as well just use that to replicate the stock behaviour more closely

https://github.com/Kopernicus/Kopernicus/commit/4374e98387f431fdf31ece6efe4868b18d2ad11a

this is a much simpler solution, and probably results in a better experience overall

PR #412, opened 2020-01-19
Update to 1.8

Sigma88:


Sigma88:
too early, need to fix more bugs

PR #413, opened 2020-01-23
Use AssetBundle.LoadFromFile rather than WWW

blowfishpro:
gets rid of deprecated code

would be nice to make this async but that's more work (and I don't really understand the implications of delaying the shader loading here)


Sigma88:
I assume this will not change the behaviour right?

blowfishpro:
Correct

Sigma88:
thanks then 👍

PR #414, opened 2020-01-23
Hazardousbody2

Sigma88:


Tuupertunut:
This is a nice solution and will probably work correctly.

PR #420, opened 2020-02-06
Re-add particles?

democat3457:


Sigma88:
I'm a bit busy at the moment so double checking this might take a while

unless @StollD  wants to take a look himself

StollD:
Nice! I totally forgot about KSPParticleEmitter. IIRC that got converted to use Shuriken Particles some time ago, but I never cared to look into it. Thank you!

However, I think we should use this opportunity to revamp the particles a bit. It should expose all the options that KSPParticleEmitter exposes, and it shouldn't wrap KSPParticleEmitter entirely, like it is done right now. Instead the Kopernicus loader should edit KSPParticleEmitter directly, and add a component that fetches KSPParticleEmitter and manages the target / timewarp stuff at runtime.

We also have to check if ParticleSystem still fetches the mesh that tells the emitter where to emit particles from a MeshFilter on the same GameObject, or if that has changed.

democat3457:
This may sound dumb, but how do I do a Gradient Parser?

democat3457:
Actually nevermind I just realized its a Min-Max Gradient

democat3457:
Do you want to expose options that are included in various ParticleSystem modules such as TrailModule and EmissionModule?

democat3457:
I wasn't sure what you meant by add a new component so I edited the existing one.

src/Kopernicus/Kopernicus.csproj
```diff
@@ -44,10 +44,6 @@
       <HintPath>..\..\packages\UnityEngine.AssetBundleModule.1.8.1\lib\net40\UnityEngine.AssetBundleModule.dll</HintPath>
       <Private>False</Private>
     </Reference>
-    <Reference Include="UnityEngine.CoreModule, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null">
```
Sigma88: you should not remove this

src/Kopernicus/Kopernicus.csproj
```diff
@@ -64,6 +60,12 @@
       <HintPath>..\..\packages\UnityEngine.UnityWebRequestWWWModule.1.8.1\lib\net40\UnityEngine.UnityWebRequestWWWModule.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <Reference Include="UnityEngine.CoreModule">
```
Sigma88: and you should not add this

src/Kopernicus/packages.config
```diff
@@ -1,11 +1,13 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="Assembly-CSharp" version="1.8.1" targetFramework="net472" />
+  <package id="Defiant_Zombie.KSP.Skeleton.UnityEngine.CoreModule" version="1.8.1" targetFramework="net472" />
```
Sigma88: you should not add this

src/Kopernicus/packages.config
```diff
@@ -1,11 +1,13 @@
...
+  <package id="Defiant_Zombie.KSP.Skeleton.UnityEngine.ParticleSystemModule" version="1.8.1" targetFramework="net472" />
```
Sigma88: @StollD  I think this should be added to your jenkins rather than doing this, right?

democat3457:
Strange, I didn't see these comments until now - anyway, I will resolve as soon as possible.

src/Kopernicus/packages.config
democat3457: This is added because it is a dependency of ParticleSystemModule.

democat3457:
Yucky merge commits but it wouldn't detect that I merged the conflict.

R-T-B:
Hello, Kopernicus is being forkek at the moment at request of it's creator.  You may want to move this pull request to the most popular branch on the forums, which I and user Presja are maintaining.

https://github.com/prestja/Kopernicus

We will be happy to consider this pull request in proper once the 1.9 update is properly finished.

R-T-B:
Heh, looks like they like your code anyways.

@Sigma88, I am correct that this Kopernicus fork is a dead end right?  And if so, on our fork, any comments as to this codes quality while we work on other things?  We do plan to try to bring it in, it's just we have bigger fish to fry right now.

StollD:
> Heh, looks like they like your code anyways.

I always welcomed the code. But I still think that old hacks should not simply get merged back in, but instead it should be done better if there is the chance to do it.

The whole `PlanetParticleEmitter` component was a huge an ugly hack. It was never the best solution, but it was neccessary to keep the jiggling of 4 or 5 different particle components out of the main config loader code.

Now, when using `KSPParticleEmitter` you only need to care about two components: The particle emitter itself, and a small component that applies the stuff that needs to be done at runtime, like directing the particles away from the sun. And you don't need a lot of setup either, you just create both components and are good to go. You dont need to duplicate all their properties, and apply them in the `Awake()` method, because you can just apply them directly through the config loader.

This would also make it possible to edit the full particle system using Kittopia. With the current changes, every option that is set by `PlanetParticleEmitter` in `Awake()` couldn't be changed later anymore, because `Awake()` only runs once, when the planet is spawned by KSP. 

Then, because `KSPParticleEmitter` is based on the new Shuriken ParticleSystem in Unity, it probably exposes a lot more options that the old legacy particles. I imagine that you could do pretty cool stuff with these options (like having a particle system emit meshes, to create something like boulders / asteroids). Instead of just setting some arbitrary default, these options should be exposed properly as config options for the planet modders to use them. Since there was already a release without the particle support at all, you don't need to really care about backwards compatibility, and don't need to keep the same config options that the old code had, like the changes in this PR do.

There are also some nitpicks that I have that don't matter too much I guess:
* You shouldn't load the shader directly from a string, because not all shaders are the same. Nothing guarantees that the shader has support for a `_MainTex` value for the texture, etc. Either hardcode the shader to one that is know-good, or use the Material wrappers that Kopernicus already uses everywhere else (PQS surface material, scatter material etc.), which has the side effect that the user can automatically edit all the properties of the shader, not just the texture
* `if (target != "None")` that is also a leftover from me writing silly code, and should probably be improved (keep the default value of `null`, and check if the body that is returned by `UBI.GetBody()` is null before using it.

Since my idea is basically "throw everything away and redo it properly", I wanted to do some of the stuff myself, but I never found the time to look into it properly, so this PR was left in limbo. And I really wanted to get out as much of the legacy stuff as I could, so I didn't want to simply merge it or close the PR without a comment. But it should work just fine, if you decide to merge it.

democat3457:
I see...

prestja:
Particle support has been added through PR #450

PR #423, opened 2020-02-23
Localization

tinygrox:
made some changes for Localization.
Is this the branch?


Sigma88:
A couple of thoughts

1- I don't know how I feel about including some languages into the localization but not others, I think maybe the other languages could go into a language pack rarther than kopernicus itself

2- I think that using Localize.Format comes with a performance hit, so I guess we should format the value once and then keep using the formatted string

tinygrox:
@Sigma88 I can only do so. And I don't quite understand what you mean by "language pack".

Sigma88:
thank you for making the changes

>I don't quite understand what you mean by "language pack".

I mean a separate download

hardcoding only one extra language would feel weird to me.

I am not necessarily against it, but I would rather defer to @StollD  on that one

src/Kopernicus/Components/KopernicusSolarPanels.cs
```diff
@@ -37,7 +38,16 @@ namespace Kopernicus.Components
     /// </summary>
     public class KopernicusSolarPanels : PartModule
     {
-        [KSPField(guiActive = true, guiActiveEditor = false, guiName = "Tracking Body", isPersistant = true)]
+        //Strings for Localization
+        string SP_status_DirectSunlight = Localizer.Format("#Kopernicus_UI_DirectSunlight");//"Direct Sunlight"
```
Sigma88: I think the new unity might have some issues with this way of defining the variables.

you should change them into static strings

I am not sure if changing them to static is enough to fix all issues or if we should also define them at a different time

StollD:
> I mean a separate download
> 
> hardcoding only one extra language would feel weird to me.

We *have* to ship at least one language or make that language pack a requirement. The language pack cannot change the texts that Kopernicus uses.

And I would imagine that more languages will follow as people contribute them, just that @tinygrox was only able to provide an english version to get started.

Sigma88:
>We have to ship at least one language or make that language pack a requirement. The language pack cannot change the texts that Kopernicus uses.

What I mean is: English will be included as the default, having just one extra language would feel weird to me

Sigma88:
hardcoding all the languages would require a new release everytime a change in the localization is required (grammar checks, new languages added, etc)

StollD:
Ah wait, there is a chinese version too. Now that makes more sense, but I don't think it is bad or weird.

If there is another language, or a fix to one of them, you can still download a random other file, put it into GameData and Kopernicus will use it. And we don't have to release a new version immediately, we can still wait until there are enough changes that warrant a release. We provide the feature, we decide how to handle it.

src/Kopernicus/Components/KopernicusSolarPanels.cs
StollD: Argh, you beat me to it. No, Unity won't have issues (using the constructor in a MonoBehaviour is totally fine as long as you don't create it with `new MB();`), but it is unneccessary to store a copy of these strings for every instance of the solar panel module.

Please make them `private static`.

Sigma88: @tinygrox just do what @stolld said and we can merge your PR

Sigma88:
I would also suggest to move the localization into:

`Kopernicus/Localization/english.cfg`
and
`Kopernicus/Localization/chinese.cfg`

tinygrox:
Is that right?

Sigma88:
Yes, thank you!

Sigma88:
Wrong button

PR #425, opened 2020-02-26
Fix loading of R8 and A8 textures and add R16.

RCrockford:
I've been trying to get R8 (or A8) heightmaps working for RSS, but they fail to load, even though they are listed as supported.
I noticed in the code that the wrong flags were being read; this PR corrects that.

I switched R8 textures to load as R8 rather than A8 as the latter doesn't work for heightmaps. I guess this might break existing uses, but given the bug in the texture loader, I suspect there aren't any.

I have also added support for R16 textures, to allow for better heightmap resolution.

### DIFF
--- a/src/Kopernicus/OnDemand/OnDemandStorage.cs
+++ b/src/Kopernicus/OnDemand/OnDemandStorage.cs
@@ -403,11 +403,11 @@ namespace Kopernicus.OnDemand
                                 new DDSHeaderDX10(binaryReader);
                             }
 
-                            Boolean alpha = (ddsHeader.dwFlags & 0x00000002) != 0;
-                            Boolean fourcc = (ddsHeader.dwFlags & 0x00000004) != 0;
-                            Boolean rgb = (ddsHeader.dwFlags & 0x00000040) != 0;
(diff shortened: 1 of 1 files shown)


Sigma88:
this looks fine to me

unless @StollD  has any objection I will merge this before releasing 1.9.0-1

Poodmund:
@RCrockford unfortunately this commit caused the loading of the Palette-4bpp and Palette-8bpp DDS formats to break and causes any planet pack using these formats (for biome maps) to not load, Kopernicus throws the exception on game load as the biome textures are null.

Example of Palette-4bpp texture: https://github.com/Poodmund/Outer-Planets-Mod/blob/master/GameData/OPM/OPM_Textures/PluginData/Polta_biome.dds

Example of Palette-8bpp texture: https://github.com/Poodmund/Outer-Planets-Mod/blob/master/GameData/OPM/OPM_Textures/PluginData/Tekto_biome.dds

Poodmund:
https://github.com/Kopernicus/Kopernicus/commit/9a5d17cfe3a993024f22e77e3ee29701e03b2e46 seems to resolve the issue.

PR #429, opened 2020-03-16
Finally we have SolarPanels

Sigma88:
Finally this does what I think most people expect from multiple stars in Kopernicus.

The override for the CalculateSunBodyFlux is mostly unchanged, I just added back in `GetAtmoThermalStats` that was missing for some reason in the previous iterations of the override.

The new behaviour of the KopernicusSolarPanels is as follows:

- there are 2 different PartModules (`KopernicusSolarPanelsFixer` and `KopernicusSolarPanels`)
- `KopernicusSolarPanelsFixer` should always be the first `MODULE` in the cfg
- `KopernicusSolarPanels` should always be the last `MODULE` in the cfg
- the `FixedUpdate` order will be
  1. `KopernicusSolarPanelsFixer`
  2. All `ModuleDeployableSolarPanel`s
  3. `KopernicusSolarPanels`
- `KopernicusSolarPanelsFixer` will change the PhysicsGlobals.Luminosity to the one of the tracked body
- `ModuleDeployableSolarPanel` will calculate tracking normally
- `KopernicusSolarPanels` will set the tracking speed to 0 and recalculate the tracking for all remaining stars

this way, we get the intended Ec from all stars and everything is just re-set to the normal values after each Part.FixedUpdate

also, if someone were to make up custom solar panels derived from MDSP they should be automatically compatible with Kopernicus


Sigma88:
Minor Note:

I made a small change in the trackingBodyName code so that now if the body is automatically selected the name will show up with a small tag after

eg: `The Sun (Auto)`

This name is localized using the key ```#Kopernicus_UI_AutoTrackingBodyName = <<1>> (Auto)``` in both english and italian

@tinygrox  what would be the best Chinese translation for that ?

Sigma88:
I hit a couple of bumps but the idea was good, just needed some tweaks to avoid stupid bugs

now it is ready to be reviewed/merged

StollD:
Sorry, I *totally* forgot about this! Too much going on otherwise...

Looks good to me. Only question I have: Why are there two solar panel modules now? Couldn't you use TimingManager to have both run within the same module?

Otherwise feel free to merge this one.

tinygrox:
> @tinygrox what would be the best Chinese translation for that ?

Sorry, I just saw this today.
`"自动"` is OK. So the key should look like `#Kopernicus_UI_AutoTrackingBodyName = <<1>> (自动)`.

Sigma88:
thanks

gotmachine:
Inserting a module at index zero is a very bad idea. 
It's gonna break all parts using a module that rely on a module index in configs : ModuleAlternator, ModuleCargoBay, ModuleDataTransmitter, ModuleSAS, ModuleScienceLab, ModuleSurfaceFX, the various ModuleWheels* variants...

Sigma88:
in what way do they rely on the index? I haven't had any issues with my testings

gotmachine:
They reference another module by their index.
ModuleAlternator.engineIndex
ModuleCargoBay.DeployModuleIndex
ModuleDataTransmitter.animationModuleIndex
ModuleSAS.CommandModuleIndex
ModuleScienceLab.containerModuleIndex
ModuleWheelBrakes.statusLightModuleIndex
ModuleWheelBogey.deployModuleIndex
ModuleSurfaceFX.thrustProviderModuleIndex

I don't think there is a case where one of these is present alongside a ModuleDeployableSolarPanel in stock, but there is definitely many cases in part mods.

Sigma88:
frick
why do people do this kind of s.....
smh

should be fairly easy to fix

Sigma88:
oh and btw, thanks for the headsup @gotmachine  👍

Sigma88:
ok, so now instead of inserting the first module at the beginning I do this:

- copy the first module solar panel at the end
- replace the first module solar panel with one Kopernicus module
- add the second Kopernicus module at the end


this should be fine right?

StollD:
> ok, so now instead of inserting the first module at the beginning I do this:
> * copy the first module solar panel at the end
> * replace the first module solar panel with one Kopernicus module
> * add the second Kopernicus module at the end
> 
> this should be fine right?

I still don't really understand why you need to use two modules. Wouldn't it work the same if you did it like that (using TimingManager): https://gist.github.com/StollD/d0e0ffe7e9ca3cb1ab5f9f49b5c6da29 ?

Sigma88:
Timing manager doesn't work properly for what we need.

We need to run code before and after each vessel while timingmanager can only run before and after all vessels
```
-TimingManager (n)
-Kopernicus
-Vessel1
-Kopernicus
-Vessel2
-Kopernicus
-TimingManager (n+1)
```

an alternative would be to just add a generic kopernicus component to every vessel during runtime and avoid all the "KopernicusSolarPanels" MM cfg headache but I went with the cfg because that was the previous accepted solution

gotmachine:
Reviewing the code quickly, some comments : 

1. You are calling `KopernicusStar.CalculateFluxAt(Vessel vessel)` **for each** solar panel module **2 times** (in each module), and one extra call for each extra star. Physics raytracing really isn't cheap, this should really be done only one time per vessel.
Morevover, since you are also calling CalculateTracking -> CalculateTrackingLOS for every star, you get 2 extra raytracing call per module and per star, and while the part/terrain LOS raytracing call is relevant, the first scaledbody LOS raytracing call is almost entirely redundant. 
But, well, that is stock behavior and harder to work around if you want to keep calling the overridable MDSP methods to have compatibility with MDSP derivatives (which IMO isn't worth the trouble, as in practice the vast majority of solar panel plugins are reimplemented from scratch and won't be compatible anyway)

2. Regarding the execution order issue, IMO it would be easier if you just decided to take control of the target module(s) execution flow :
- In `OnStart`, find all MDSP and derivatives on the part, save a reference to them, and set `enabled` to false. This will prevent the module MonoBehaviour methods (ie, `FixedUpdate`) from being called, but won't alter the KSP methods and lifecycle, in particular `OnStart`, `OnLoad` and `OnSave` (KSP only check for `isEnabled`, which is PartModule specific and is unaffected by the state of `enabled`). 
- From your partmodule, for each found MDSP, call every implemented MonoBehaviour method in MDSP from your own implementation : `FixedUpdate`, `LateUpdate`, `OnCollisionEnter`. Ideally, if you want to be on the safe side to support a MDSP derivative, you can reflection find/call the most common unity methods that aren't implemented in MDSP, especially `Update` and `Start`.

Sigma88:
this might be a good idea. at the moment I am busy with other stuff but if I get time I will try to go back and sort this mess out

Sigma88:
@prestja since you have a released version of kopernicus you might be inteterested into adding this feature.

I ended up not including it because of the comments in this pr but since a version of kopernicus is out, it might as well have this feature included.

R-T-B:
I have advised him of this branch.  He has invited me to contribute to Kopernicus development and we are already looking at it.

Besides the above, any "heads up" in here we should be aware of?

PR #506, opened 2021-12-02
Update README.md to fix a name

teknoman117:
Fix my name


R-T-B:
I see no reason why not.  If you see anything else in the codebase that needs fixing, let me know.

PR #532, opened 2022-07-16
Improve SunBodyFlux performance

JonnyOThan:
-cache the home body SMA for purposes of SolarLuminanceAtHome
-cache the star name instead of using Object.name which allocates a string for every call
-don't use reflection to set PhysicsGlobals.solarLuminosity

-this results in a ~10x speedup of FlightIntegrator.LateUpdate in a Galaxies Unbound install with all optional systems


R-T-B:
This is actually a surprising amount simpler than I thought it would be, and the results speak for themselves.  Rather than forcing this one through the bleeding edge, given the changes are not exactly looking like they'll cause trouble, I'm just going to mainline this now.  Thanks for the work.

PR #533, opened 2022-07-17
Fix KopernicusSunFlare.LateUpdate performance

JonnyOThan:
this was slow because it iterates over all map targets (bodies + vessels) and runs for each star in the universe

This was modeled after the stock code, which excludes bodies with localscale <1 or >3.  The scaledspace bodies seem to have a neutral scale at 1000 radius units - that means Jool is scale 1 and everything else in the stock system is smaller, and most bodies would just be ignored by this code anyway.  The effect of the stock code is to immediately turn off the sun flare when the sun goes behind jool's horizon - but the lensflares still turn off for other bodies, they just fade out.  Deleting this code causes large bodies to have the same behavior as the smaller ones (fading out) and massively improves performance for systems with large numbers of stars and bodies.

### DIFF
--- a/src/Kopernicus/Components/KopernicusSunFlare.cs
+++ b/src/Kopernicus/Components/KopernicusSunFlare.cs
@@ -1,4 +1,4 @@
-/**
+﻿/**
  * Kopernicus Planetary System Modifier
  * ------------------------------------------------------------- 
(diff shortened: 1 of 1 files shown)


R-T-B:
looks good at least in concept,  Will test in gameplay, and then release in next stable build if all goes well.

PR #534, opened 2022-07-17
Fix RuntimeUtility.LateUpdate performance

JonnyOThan:
The bulk of the cost here was KopernicusStar.GetBrightest, which called KopernicusStar.GetNearest.  GetNearest was slow and incorrect anyway (it returns the farthest star) so it's been deleted.

LateUpdate is still O(bodies_with_atmosphere * number_of_stars) because of AtmosphereLightPatch.  It's fast enough for now, even in GU, but for systems with lots more bodies and stars it might need to be revisited.

### DIFF
--- a/src/Kopernicus/Components/KopernicusStar.cs
+++ b/src/Kopernicus/Components/KopernicusStar.cs
@@ -101,40 +101,21 @@ namespace Kopernicus.Components
         public string StarName;
 
         /// <summary>
-        /// Returns the star the given body orbits
-        /// </summary>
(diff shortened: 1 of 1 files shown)


R-T-B:
"GetNearest was slow and incorrect anyway (it returns the farthest star) so it's been deleted."

Woah.  I have no idea how that slipped by.  Bugs in plain sight!  Thanks for this.  This is all good finds in code I was 99% sure was bug-free, and may explain a bug I was recently dealing with!  Yay!

PR #535, opened 2022-07-17
Improve KopernicusOrbitRendererUpdater performance

JonnyOThan:
-previously this was calling through reflection which caused GC allocs every frame
-since this was called per body, the cost added up pretty significantly in systems with large numbers of bodies

This change takes the cost in GU from 0.81ms/frame to 0.16ms/frame, or about 5x faster


R-T-B:
looks good to me.

PR #557, opened 2022-12-11
Update ring shader for separate backlit appearance

warriorsabe:
Linux shaders are currently not updated, as I couldn't get them compiled. Should look identical to before when used with old configs, and makes minimal changes to the code - simply reusing what capability already existed, but was not utilized.


R-T-B:
I finally got a hold of a mac to test the linux shaders on, hopefully they will just work as is.  Stand by.

R-T-B:
So the mac shaders work fine on mac, but the mac shaders being used as linux shaders have issues (rings always black).  I wonder, what was added by this?  Are we sure the old dual-camera ready linux shaders did not work as is?  Do you have a test case for your new functionality that I could test them against?

R-T-B:
scratch all of the above, in a strange miracle, the Windows shader bundles work perfect on both mac and Windows.

That will provide all we need.  Bizzare, but I'm not complaining!

Expect to have this merged into a new release by end of the week.

PR #583, opened 2023-04-10
Fix memory leak from Injector not removing event handler

JonnyOThan:


R-T-B:
You tested this, and it works?

Injector is a weird class, it's where all the "magic happens" (really it's just a garbage collection mess), but at any rate, I remember some former devs (TMSP?) advising me to leave that in when I fixed the other memleaks because they thought it may be needed for some reason or another.

But if it works, and is tested, I'd say there is no argument from me to get rid of it.  I'll merge as soon as I can test, or get word from you or another.

JonnyOThan:
Huh, no in fact I haven't tested it!  But these kinds of errors are common in all sorts of mods, the fixes are trivial and the effects can be significant.  I'm curious why they suggested keeping it; as far as I can tell the event will only ever be fired once per game session.

R-T-B:
I'm not entirely sure myself but that's the only reason it's still there.  I'll take a test drive tomorrow and we'll find out for sure.

R-T-B:
Sorry for the delay with this, Timberborn had a game update and all my mods there broke.

I'll be testing this in the near future though.

R-T-B:
I'm done with Timberborn (for now) and switching back to KSP with my mod sweep.  This seems harmless in my limited testing, will merge for now, and I expect no hidden issues.  Sometimes "wisdom of the ancients" is a bad approach to deving, lol.

PR #595, opened 2023-08-06
Update SolarPanels.cfg for compatibility with Kerbalism

themaster402:
Disable Kopernicus multistar when Kerbalism is detected.

Kerbalism does not recognize the KopernicusSolarPanels module as a solar generating source, and thus patched solar panels will not work with Kerbalism's background processing; as Kerbalism has its own multistar support, it would be simplest to disable the Kopernicus module.

### DIFF
--- a/build/KSP19PLUS/GameData/Kopernicus/Config/SolarPanels.cfg
+++ b/build/KSP19PLUS/GameData/Kopernicus/Config/SolarPanels.cfg
@@ -1,11 +1,11 @@
 // If any modder adds useKopernicusSolarPanels = false to a module instead of a part, add it to the part:
-@PART:HAS[@MODULE:HAS[#useKopernicusSolarPanels[?alse]]]:FINAL
+@PART:HAS[@MODULE:HAS[#useKopernicusSolarPanels[?alse]]]:NEEDS[!Kerbalism]:FINAL
 {
 	%useKopernicusSolarPanels = false
 }
 
 // Uses regular expressions to convert any case variants like FalSe to false
-@PART:HAS[#useKopernicusSolarPanels[*]]:FINAL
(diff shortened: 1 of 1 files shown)


R-T-B:
I thought kerbalism already did this here, using the old module disabler that should still work?

https://github.com/Kerbalism/Kerbalism/blob/master/GameData/KerbalismConfig/Support/Kopernicus.cfg

If that is not working, determining why should be our first priority, as a few mods rely on this mechanism.  Advice welcome, I am notoriously bad at MM configs.

R-T-B:
Just tested it in my live save and background solar processing is working with Kerbalism default config, at least.  Are you doing anything unusual here?  I do want to make sure to cover all scenarios.

themaster402:
> I thought kerbalism already did this here, using the old module disabler that should still work?
> 
> https://github.com/Kerbalism/Kerbalism/blob/master/GameData/KerbalismConfig/Support/Kopernicus.cfg
> 
> If that is not working, determining why should be our first priority, as a few mods rely on this mechanism. Advice welcome, I am notoriously bad at MM configs.
> 
> I use kerbalism in my save, a moment why I validate if this is a real issue. I swear I should've noticed it by now, but maybe I am just oblivious lol.

So upon closer inspection the issue here looks to be that [OtherWorlds Reboot](https://github.com/pkmniako/OtherWorlds_Public/blob/main/gamedata/OtherWorldsReboot/ConfigurationFiles/MultiStarSolarPanels.cfg) includes its own copy of the Kopernicus multistar config that ends up overriding the Kerbalism patch. The behavior of the Kerbalism patch is as-expected with OWR removed.

image
_With OWR_
image
_OWR removed_

Is it intended for a copy of the multistar solars patch to be included in star system mods? I'll probably notify the OWR maintainer of the issue.

Sorry for the bother! 🙂

PR #601, opened 2023-10-22
Fast biome query

gotmachine:
Port of the [KSPCF FastBiomeQuery patch](https://github.com/KSPModdingLibs/KSPCommunityFixes/pull/151) (look there for further details on what this does and the implementation details) for Kopernicus :
- Less memory usage : uses 1 bpp biome index lookup tables instead of 3bpp RGB textures
- For biome querying, on average 15-20x faster than the stock implementation, up to 30-40x for bodies with a large amount of biomes
- Fixed stock bug in bilinear interpolation on biome transitions
- Adds a small upfront texture to lookup table conversion cost, 0.5s in total for stock+OPM.

Related changes :
- The "BiomeSampler" cache has been removed, since with those changes it has become (vastly) slower than just querying the biome directly. Its public API is left exposed to avoid Parallax from breaking, but marked obsolete.
- Regression : editing biomes in Kittopia isn't applied in-game anymore (but they will be exported as before)
- Removed OnDemand for biome maps. Rationale :
  - Can't find any planet pack using it anyway
  - Less relevant since biome map memory usage is reduced by 1/3
  - Would cause significant stutter because conversion from the texture is costly

I've done some testing with RSS and OPM, as well as Stock + Parallax and things seems to work as expected, but I would recommend further testing on real installs before pushing this. 

There is some small potential of breaking mods if they do some sketchy custom biome querying instead of using the stock API. I already fixed such an issue in Kerbalism (pushed the fix in Kerbalism 3.18), but what Kerbalism did previously was likely already breaking in some corner cases.

I will push the same optimization to a future KSPCF release. The KSPCF implementation runs after Kopernicus has done its own thing and is able to detect if Koperncius already optimized the biome maps, so both mods won't step on each other feets.


R-T-B:
Thanks for this got.  I'm basically the sole tester available from our limited team at the moment, and I'm fighting off the tail end of a covid infection.  As such, it will be tested but may take a bit.  I don't trust such a major change to the bleeding edge solely because they seemingly don't get tested very well (last BE release set let several bugs through), so I'll be doing this one personally.  I'd say expect as little as a week to as long as a month to merge.

R-T-B:
I am officially begining testing of this on my savegame today.  I should have a conclusion within the week.

Sorry for the holdup, work backlog from COVID was severe.

gotmachine:
No problem, I don't see any reason to rush this.
Hope you're finally recovering and well :)

R-T-B:
Yep, I'm well now, work life is still recovering lol.

Finally got a good gaming session in.  None of the bugs I saw before are present (with the new Kerbalism build of course).  No bugs of note yet and performance seems good, I even did a complex mun landing (dock at orbital station hub and deploy a small lander shuttle to Mun surface)

A good start.  Will continue to test.

R-T-B:
It passes muster in my testing.  I'll push this to a release as soon as I get the docs updated for that.  You'll be getting credit of course.  Your contributions to this project have been invaluable.

PR #604, opened 2023-11-16
Fix an issue with the new biome code and cbNameLater

NathanKell:
Previously the old bodyName would work in FlightGlobals.GetBodyByName (which mods like RP-1 and Parallax relied on, when RSS was present) because an old bit of fast-biome code called FlightGlobals.GetBodyByName before the bodyName got changed. Now that is no longer true so FlightGlobals's bodyNames cache no longer has the old body name(s) present.

### DIFF
--- a/src/Kopernicus/Components/NameChanger.cs
+++ b/src/Kopernicus/Components/NameChanger.cs
@@ -47,6 +47,10 @@ namespace Kopernicus.Components
             {
                 OrbitRendererData data = PSystemManager.OrbitRendererDataCache[b];
                 PSystemManager.OrbitRendererDataCache.Remove(b);
+                // Before we make the change, we need to make sure FlightGlobals's bodyNames cache
+                // has the old name (for back-compat with older mods)
+                FlightGlobals.GetBodyByName(b.bodyName);
+                // Now we can change the name safely.
                 b.bodyName = newName;
                 PlanetariumCamera.fetch.targets.Find(t => t.name == oldName).name = newName;
                 data.cb = b;


NathanKell:
cc @gotmachine

gotmachine:
Nice, thanks for getting to the bottom of this.

NathanKell:
> I would maybe suggest to put both the old and new name in the FlightGlobals.bodyNames dictionary when cbNameLater is used (by calling FlightGlobals.GetBodyByName() twice in NameChanger.Start(), before and after the name change). But maybe that's a bad idea that will break some random plugin relying on the current inconsistent behavior....

Sorry I missed this earlier! That's precisely what is already happening, because the instant someone calls FindBodyByName with the new name, it will indeed get cached. Unless what you mean is "there will be mods that reflect into the dict, so best populate it ourselves then"? Because that cache is not otherwise helpful, and iterating through FlightGlobals.bodies will only ever give you the new name; the starting name is only preserved (a) in the gameobject name (but not in CelestialBody.name), and (b) as an entry in the dict.

I may be totally be misunderstanding what you're saying, but just to throw some stuff at the wall to clarify just in case:

All GetBodyByName does is:
* Look up to see if we have a cached link between the string arg and a CB. If so return that.
* if not, go through FlightGlobal's bodies list testing the string arg against the CB's name. If found, add the pair (arg, CB) to the dict [note, not replacing any other string->that_CB links!] and return the CB.
* Return null if no success.

It has nothing to do with displayName (that's just for UI display, and mods prior to the loc update, and probably some after it too knowing mods...) will just print .name instead anyway.


The situation this PR creates, which is the situation that existed before the biome refactor, is, for any body using cbNameLater, you can find it with GetBodyByName using either its original name or its nameLater name.

PR #614, opened 2024-03-15
Fix singletons

JonnyOThan:
These are generally not serious issues but it helps keep the memory reports cleaner.


JonnyOThan:
OK never mind, apparently the startup scenes is not a bitflag but this still compiles for some reason

R-T-B:
After a bit of testing in my personal save, I can finally say it looks good.  Will be merged.

Sorry this took so long btw, Helldivers 2 bug got me.

R-T-B:
having some issues with githubs merge button, not sure if site issue or something else.  Will check again in the morning.

PR #618, opened 2024-05-14
Allow modification of ring shadow radius

warriorsabe:
Just provides a config-facing radius multiplier value that is multiplied with the planet radius as its being passed to the ring shader so that planet pack authors can scale the shadows to match things like oblate objects


R-T-B:
I tested this and see no issues with it.  Merging.

PR #619, opened 2024-05-24
Various refactorings and bugfixes

gotmachine:
- Refactored the scatters lethalRadius feature to have less overhead when used, and no overhead when not used.
- Restored the scatters HeatEmitter thing. I have some doubts the base implementation for it works, and the performance overhead is atrocious, but well, saying no to planet modders is never an option.
- Fixed some nullrefs in the PQSLandControlFixer thing, refactored as a cleaner harmony patch instead of as an additional PQSMod spammed on every body.
- Fixed some nullrefs in the sentinel contracts remover, cleaned up the implementation
- Fixed some potential nullrefs in SinkingBodyFix that could lead to colliders not being restored when a body doesn't have a scaled space collider. Not entirely sure this was the issue, I couldn't actually reproduce it.


R-T-B:
Merging.  While I'm sure it's all positive changes, I'm going to still test it for a day with my test team and personal save.  Learned the hard way to do that, as I'm sure you're aware.

Will most likely ship in a day or two.  Thank you again.

gotmachine:
You might want to wait a bit, I'm not entirely sure the `PQSLandControlFixer` refactor is doing what it should.
Of course, that would help if I actually knew what this "fix" is actually trying to fix. I have some doubt this is doing anything useful.
I will try to check the git history in search for a clue, but I'm not hopeful.

R-T-B:
Yes, it was originally intended to prevent flicker, but I have doubts it is doing anything anymore as well.

I think we can remove it for now, so I'll merge these changes.  I tested them and didn't notice any flicker.  But we should be prepared to readd it, just in case.

R-T-B:
https://forum.kerbalspaceprogram.com/topic/200143-112x-kopernicus-stable-branch-last-updated-may-25th-2024/?do=findComment&comment=4392990

Could this be related?  If so, what do we think is happening?  I'm fairly at a loss as I don't think this was ever meant to treat "terrain brightness."

It could just be the user, of course.  Just marking this to keep track of it.

Issue #592, opened 2023-07-24
Memory allocations


JonnyOThan:
Can anything be done about these?
![image](https://github.com/Kopernicus/Kopernicus/assets/9804078/b7c0b0a6-6ebf-4bc5-ba6c-7ca83f22489b)

![image](https://github.com/Kopernicus/Kopernicus/assets/9804078/6159d2b0-06f0-45eb-8126-cbf05142cfa0)

![image](https://github.com/Kopernicus/Kopernicus/assets/9804078/e3827118-fc4e-4e90-8fe7-c142bf582d40)

R-T-B:
I was actually just looking at the RuntimeUtility 9nes when I switched to Rider in my Linux transition.  It flagged them for me as performance pigs, so yes those are being looked at at least

Unfortunately due to how the StarlightSwitcher works it just is going to be kinda expensive.  We have some options though, ranging from looking for low hanging optimization fruit to completely eliminating that system in "single star" mode.  tl;dr: I don't have a timeframe yet but this is on my radar.

PR #620, opened 2024-05-24
Remove PQSLandControlFixer + GC/perf fixes

gotmachine:
- Fixes for issue #592

- PQSLandControlFixer was a remnant of an obsolete fix, that hasn't been doing anything useful since release 155
See https://github.com/Kopernicus/Kopernicus/pull/619#issuecomment-2129343926


R-T-B:
Tested out fine in my day long test run, we will be pushing this.

Issue #633, opened 2024-07-26
Problems with SolarPanels compatibility in multi-star mode


Aebestach:
There are two special solar panel modules in NearFutureSolar and UniversalStorage2, ModuleCurvedSolarPanel and USSolarSwitch, which are not replaced by KopernicusSolarPanel, so they cannot correctly identify the sun in multi-star mode.

R-T-B:
Can't believe this has not been reported before, certainly needs fixing.  It looks like we may need to code a custom module since these are not based on ModuleDeployableSolarPanel like most modules.  I'll look into this when I next get some time to work.

Aebestach:
I also found this problem when using WeatherDrivenSolarPanel. Perhaps KOP can refer to [kerbalism](https://github.com/Kerbalism/Kerbalism/blob/f5cd99c13274599f9f8964dc162cc887e4ac85ac/src/Kerbalism/Modules/SolarPanelFixer.cs#L32) to solve this problem.

R-T-B:
Kerbalism uses its own multistar SolarPanel logic IIRC.  It won't be of much use here because of deep integration with Kerbalism itself.

Aebestach:
If I understand correctly, kerbalism is a separate solar panel power generation model, which has an abstract class SupportedPanel to support all solar panel modules including ModuleDeployableSolarPanel.
https://github.com/Kerbalism/Kerbalism/blob/f5cd99c13274599f9f8964dc162cc887e4ac85ac/src/Kerbalism/Modules/SolarPanelFixer.cs#L841

So can KOP also be written in this way? If this is to be achieved, the entire KopernicusSolarPanel code may need to be _rewritten_.

I am curious to ask how you think, I would like to learn from it and use it in WDSP. (But everything will wait until you have time)

R-T-B:
We did a cost analysis and actually, there are only really two problematic modules I can find, ModuleCurvedSolarPanel  in NFS, and of course the US2 module you listed which is based on the same code.  Supporting both with the same module should be fairly trivial.  As such, to save time, I am just going to implement support for these directly rather than doing a whole rewrite.  Then we can focus on other feature requests (some of which I'm sure you know about lol) in the coming month.

R-T-B:
Test build for this issue and NearFutureSolar:
[Kopernicus-1.12.1-207dev.zip](https://github.com/user-attachments/files/16397465/Kopernicus-1.12.1-207dev.zip)

[...]

PR #636, opened 2024-08-03
Rewrote the KopernicusSolarPanel class

Aebestach:
Rewrote the KopernicusSolarPanel class. To fix issue #633 . 
An abstract class is provided that can now support multiple solar panel types. Except for US2, all other known modules involving solar panels have been adapted.

**However**, the new KopernicusSolarPanel no longer uses the calculation method of the Stock version in single-star mode, so it may need to be modified.It may not need to be modified, and the calculation mode is similar to the Stock version.


Aebestach:
I am considering whether to add an analysis status

R-T-B:
Oh I'm sorry, I thought you wanted to move this into Master branch lol.

Feel free to remake it targeting that.

PR #637, opened 2024-08-09
Fixed various bugs in multi-star mode and some other issues

Aebestach:
1. The bug that EC still generates when switching tracking targets has been fixed. This time it is smoother.
2. Change the automatic tracking program from the maximum brightness of all stars to the maximum brightness of the visible stars or automatically select the nearest star
3. Change the EC display unit to display three decimal places
4. The occlusion factor returned by the static board is adjusted to 0 or 1


R-T-B (approved):
looks good to me.

PR #646, opened 2024-10-10
Fix #644: Performance issue with Utility.GetMod

### LINKED ISSUE #644: Performance: Utility.GetMod should probably cache its results
I was looking through someone's profile capture and their FlightIntegrator, FlightInput, and Camera updates were very expensive.  A deep profile capture shows that it's dominated by Utility.GetMod, invoked by VertexHeightOblate:

image

https://github.com/jamespglaze/VertexHeightOblateAdvanced/blob/7263707846db6415349d5f7eeecf5a6d3d4055d2/VertexHeightOblateAdvanced/CameraHijacker.cs#L55

I don't know what else might have this problem but it seems like maybe Kopernicus should just cache the results for each body assuming they never change.

https://github.com/jamespglaze/VertexHeightOblateAdvanced/issues/3

JonnyOThan:
-VertexOblateHeight calls this many times every frame, and the old implementation does a full traversal of the PQS transforms (can be thousands of game objects) 
-The PQS already has a list of mods (though it's private).  Just use that instead.

### DIFF
--- a/src/Kopernicus/Configuration/OceanLoader.cs
+++ b/src/Kopernicus/Configuration/OceanLoader.cs
@@ -333,7 +333,7 @@ namespace Kopernicus.Configuration
             }
 
             // Load existing mods
-            PQSMod[] mods = Utility.GetMods<PQSMod>(Value);
+            PQSMod[] mods = Utility.GetMods(Value);
             for (Int32 i = 0; i < mods.Length; i++)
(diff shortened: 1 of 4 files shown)


R-T-B:
I see no issues with this at least in concept that private list should do everything we need.  This will perform so much better, too.

PR #706, opened 2025-11-11
perf: Optimize `KopernicusCBAttributeMapSO.CompileToTexture`

Phantomical:
In my profiles this one method accounts for 437ms during scene switch. With this commit it now takes ~190ms.

The specific changes made here are:
* Replace Math.Round call with an inlinable method that should be much faster over the target domain.
* Use unsafe code and pointers to avoid boundary check and function overhead associated with indexing into arrays and native arrays.
* Use the same trick I'm using in https://github.com/Gameslinx/Parallax-Continued/pull/79 to avoid texture initialization overhead when creating a new texture.

The new actual performance will vary quite a bit based on whether the internal texture copy that happens as part of `GetRawTextureData` ends up causing a bunch of page faults or not. It may be quite a bit faster than 190ms if it ends up reusing a previous texture allocation.

Before:
image

After:
image

### DIFF
--- a/src/Kopernicus/Components/KopernicusCBAttributeMapSO.cs
+++ b/src/Kopernicus/Components/KopernicusCBAttributeMapSO.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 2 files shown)


R-T-B:
This has past testing and will be in a release soon.

PR #710, opened 2025-11-30
perf: Another pass at optimizing `KopernicusCBAttributeMapSO.CompileRGB`

Phantomical:
I have been doing some work on Sol. This method takes about 0.2s for each scene switch in my original test setup, but with Sol it takes 2.8s (even with the last round of optimization). This is another round of optimization for CompileRGB and CompileRGBA.

In order, the 3 commits here do:
* b8bea7c7bf6fa4eb0afc3e1e5233e9419922090b - A straightforward optimization where I have hoisted the calls to `RoundToPixelValue` outside of the loop into a lookup table.
* fcd27eb62ff84247c6beea37e9edd29c742539a3 - The computation is then moved to a job so it can be run in parallel with `Texture2D.GetRawTextureData`
* c902ffc7faf7284e6d0ca4a6b6361a418377e671 - This then copies the simple optimization over to CompileRGBA.

With these optimizations the cost of calling `CompileRGB` is now basically just the cost of calling `GetRawTextureData`.

### Parallax-Continued + Kopernicus
Before
image

After
image

Timings are pretty inconsistent with smaller textures. Parallelization speeds things up either way, but overall timings really depend on the memory allocator.

### Sol + Parallax-Continued + Kopernicus
Before
image

After
image

Timeline view of after
image

The overall improvement is about 1s for sol.

### DIFF
--- a/src/Kopernicus/Components/KopernicusCBAttributeMapSO.cs
+++ b/src/Kopernicus/Components/KopernicusCBAttributeMapSO.cs
@@ -567,7 +567,7 @@ namespace Kopernicus.Components
 
             // By just allocating 256 entries we don't need to worry about
             // possibly accessing out of bounds.
-            RGB* attributeColors = stackalloc RGB[byte.MaxValue];
+            RGB* attributeColors = stackalloc RGB[256];
             for (int i = 0; i < Attributes.Length; ++i)
             {
                 Color pixelColor = Attributes[i].mapColor;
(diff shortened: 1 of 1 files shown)


R-T-B:
I'll be merging this but it won't be in a release until properly tested, probably December as mentioned. But makes it easier to make a test build at least.  Same was done with a few other PRs, to facilitate widespread testing of new features.

PR #711, opened 2025-12-08
Add support for asset bundles and (somewhat) async texture loading

Phantomical:
This intended to remain as a draft PR for a bit. I do intend to fully raise this as a PR once KSPTextureLoader gets up on CKAN (should be in the next few days) but I'm hoping I can get some folks trying this out as beta testers so I can get some solid data on whether this works as expected.

This replaces the on-demand texture loading code with [KSPTextureLoader](https://github.com/Phantomical/KSPTextureLoader). I have taken care to ensure that all formats that kopernicus supports (even the undocumented palette ones) should continue to work, though this still needs more testing.

Aright, so, features:
* KSPTextureLoader has a system where it can transparently load textures from asset bundles. Kopernicus gets this for free.
* Kopernicus will now share loaded textures with other mods that happen to be loading those textures.
* Actual asynchronous loading of textures. The current PR doesn't really take advantage of this.
* Performance. Even when doing mostly sync loading this PR should be about 1.5-2x faster when loading textures.

The real benefit here is asset bundle support. This way mods can ship with all their planet-related textures in asset bundles instead of having to worry about only shipping some of them.

## Implementation Details
This PR is pretty straightforward: it swaps out the internals of `MapSODemand` to use `TextureHandle<Texture2D>`. We use `TextureLoader.LoadTexture<Texture2D>` to load the texture, grab an actual reference to the texture in `Load`, and `Dispose` of it in `Unload`. In addition, I've added `Preload` to so that texture loads can be kicked off in advance.

Everywhere else now uses `LoadTexture` to load textures synchronously.

### DIFF
--- a/src/Kopernicus/Components/KopernicusCBAttributeMapSO.cs
+++ b/src/Kopernicus/Components/KopernicusCBAttributeMapSO.cs
@@ -186,42 +186,103 @@ namespace Kopernicus.Components
             for (int i = biomeCount; i-- > 0;)
                 biomeColors[i] = Attributes[i].mapColor;
 
-            int badPixelsCount = 0;
             int size = _height * _width;
 
             Color32[] colorData = tex.GetPixels32();
             _data = new byte[size];
-
-            Parallel.For(0, _width, x =>
+            
+            var shared = new ConvertFromTextureJob.SharedData();
+            var job = new ConvertFromTextureJob
             {
-                for (int y = _height; y-- > 0;)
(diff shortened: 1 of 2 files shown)


Phantomical:
I think this is good to go. I think the one last thing to decide is: do you want me to embed a copy of `KSPTextureLoader.dll` in the repo or is it fine to assume that anyone doing development work on Kopernicus should have it installed?

PR #717, opened 2025-12-23
Change various MapSO parsers to parse as the correct type

Phantomical:
These were accidentally changed to be all RGBA instead of the correct types, this fixes that.

I have verified that this fixes the issue with MKRI but haven't tested it more broadly.


R-T-B:
I trust you guys here, will merge and push out a hotfix (pending discussion).

PR #723, opened 2025-12-28
Rework MapSODemand internals

Phantomical:
The implementation of MapSODemand has been bugging me since I have started working on it. It worked ok in the past, but with R16 textures entering the picture now it doesn't quite make sense anymore. This is a rewrite that (IMO) makes things quite a bit cleaner.

The main changes are:
* The memory format is now an enum, instead of being implicitly decided by bytes-per-pixel.
* Where reasonable (i.e. for R8/A8/R16/RA16) textures are copied over in their original format instead of always going through GetPixels32. This also means that R16 heightmaps are copied over as-is.
* R16 textures now have full support, though only `GetPixelFloat` and `GetPixelColor` are actually capable of representing the full 16-bit precision. For compatibility R16 is treated as R8A8 for HeightAlpha maps.

Otherwise:
* I have simplified the if (!IsLoaded) Load() dance into a helper function.
* The various CompileToTextures functions seemed to be rather broken, these now just call into the appropriate GetPixel* function. I don't think these end up getting used that much anyhow.

I have tested this a bit with Sol, both R8 and R16 heightmaps seem to work fine and after one bugfix colors work properly. It could likely use some more testing so I'll probably teleport around the OPM planets when I have time either tomorrow or the day after to stress test it.

cc @ballisticfox You'll probably find this PR useful for the R16 support.

### DIFF
--- a/src/Kopernicus/Components/NativeByteArray.cs
+++ b/src/Kopernicus/Components/NativeByteArray.cs
@@ -25,6 +25,7 @@
 
 using System;
 using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
 namespace Kopernicus.Components
@@ -54,6 +55,8 @@ namespace Kopernicus.Components
 
             // Allocate a pointer to the custom array
             _items = Marshal.AllocHGlobal(Marshal.SizeOf(typeof(Byte)) * Size);
+            if (_items == IntPtr.Zero)
+                throw new OutOfMemoryException("failed to allocate native array");
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


ballisticfox:
I’m pretty sure the real wrapping is handled in the ConstructBilinearCoords method

Phantomical:
I don't think so?
```cs
    protected void ConstructBilinearCoords(float x, float y)
    {
        x = Mathf.Abs(x - Mathf.Floor(x));
        y = Mathf.Abs(y - Mathf.Floor(y));
        centerX = x * (float)_width;
        minX = Mathf.FloorToInt(centerX);
        maxX = Mathf.CeilToInt(centerX);
        midX = centerX - (float)minX;
        if (maxX == _width)
        {
            maxX = 0;
        }
        centerY = y * (float)_height;
        minY = Mathf.FloorToInt(centerY);
        maxY = Mathf.CeilToInt(centerY);
        midY = centerY - (float)minY;
        if (maxY == _height)
        {
            maxY = 0;
        }
    }
 ```

edit: actually nvm, it's done by the `Mathf.Abs(x - Mathf.Floor(x))` parts

R-T-B:
Merging for testing.

PR #730, opened 2025-12-30
Modernize the build system

Phantomical:
I was looking into what would be needed to pull `Kopernicus.Parser` in, saw that it was already using an old version of KSPBuildTools, and figured that I would be better off just modernizing the build system here first.

What this changes:
* The release number is now set in `Kopernicus.csproj` and gets automatically propagated everywhere else based on that (both in the code and the version file).
* Source files are now added implicitly. I have manually excluded those in the ShadowMan unity project.
* You can now just do a build with `dotnet build` and don't have to do one through visual studio first. Building through visual studio still works just fine though.
* Harmony is now vendored under `dependencies`.

What this should keep the same:
* All the assembly metadata should remain the same.
* This shouldn't break any workflows. You can still start KSP from within vs, and the binaries are still copied over to the correct folder under `build`.
* All the existing configurations still work. I have even fixed a miscompilation for 1.8, so I'm not entirely sure if it has been used anytime in the last few years.

### DIFF
--- a/build/KSP18/GameData/Kopernicus/Plugins/Kopernicus.version
+++ b/build/KSP18/GameData/Kopernicus/Plugins/Kopernicus.version
@@ -1,10 +1,25 @@
 {
-  "NAME": "Kopernicus",
-  "DOWNLOAD": "https://github.com/Kopernicus/Kopernicus/releases/latest",
-  "CHANGE_LOG_URL": "https://raw.githubusercontent.com/Kopernicus/Kopernicus/master/README.md",
-  "VERSION": "1.8.1.234",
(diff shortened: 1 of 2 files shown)


R-T-B:
"All the existing configurations still work. I have even fixed a miscompilation for 1.8, so I'm not entirely sure if it has been used anytime in the last few years."

Not for around 2 years, yeah.  Anyhow, we plan a release sometime next week, so you may want to get your merges in for testing before the weekend is over.  Thanks for all your efforts, and obviously merging this.

Phantomical:
TBH everything else I have going is fine to be kicked to the next release. I'm not really blocked on any of these changes.

PR #734, opened 2026-01-01
Burst-accelerate MapSODemand compilation (v2)

Phantomical:
This fixes the issues with #729, and seems to work when tested with Sol.


R-T-B:
I trust you here, lol.  I'll merge for testing...

PR #735, opened 2026-01-01
Allow MapSODemand to use raw texture data when not compiling

Phantomical:
In some cases we already support the texture format directly, in these cases we can call GetRawTextureData and use the texture data directly. This is also necessary to make sure the behaviour for HeightAlpha R16 textures remains the same no matter the value of `UseManualMemoryManagement` .

I have done some light testing of this with Sol and it seems to work fine. It is also completely off the default path, so it shouldn't cause any visible issues unless somebody is doing something weird.

To try this out drop this patch somewhere in GameData
```c
@Kopernicus:FINAL
{
    %useManualMemoryManagement = false
}
```


R-T-B:
again, trust your judgement here.

PR #737, opened 2026-01-03
Burst-compile KopernicusCBAttributeMapSO.CompileRGBJob

Phantomical:
This doesn't noticeably improve load times because the limiting thing for performance is calling `GetRawTextureData` but using less CPU isn't necessarily a bad thing.

I do have something I want to try in the future here, having these jobs be burst-compiled should help if it works out.


R-T-B:
Less CPU seems good to me.

PR #741, opened 2026-01-05
Fix a couple of issues with R16 MapSODemand

Phantomical:
There were two issues here:
- `x + y << 8` is not the same as `x + (y << 8)`.
- `_bpp` was being set wrongly for R16 (and other) textures.

All the existing formats happened to now hit these issues because nobody was using the raw formats other than R8, R8A8, RGB24, and RGBA32 all of which happened to work fine when used with the correct map depth.


R-T-B:
I assume you understand whats going on here plenty well better than me, so trusting your judgement.

PR #759, opened 2026-01-18
Fix state switching

Aebestach:
1. In previous versions, being in shadow or obscured by terrain was considered invisible. This PR submission refines the definition of invisible to include being in shadow and being obscured by terrain. Additionally, a state description has been added for solar eclipses;
2. Move `GetLargeBodies()` outside the loop to avoid recalculating for each star;
3. Remove some unused code inherited from Kerbalism.

### DIFF
--- a/src/Kopernicus/Components/KopernicusSolarPanel.cs
+++ b/src/Kopernicus/Components/KopernicusSolarPanel.cs
@@ -439,13 +439,12 @@ namespace Kopernicus.Components
 
             Vector3d direction;
             double distance;
+            List<CelestialBody> occludingBodies = GetLargeBodies(position);
             for (Int32 s = 0; s < KopernicusStar.Stars.Count; s++)
             {
                 KopernicusStar starL = KopernicusStar.Stars[s];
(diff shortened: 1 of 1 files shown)


src/Kopernicus/Components/KopernicusSolarPanel.cs
```diff
@@ -1102,25 +1127,22 @@ public override double GetOccludedFactor(Vector3d sunDir, out string occludingPa
 
                         occludingPart = blockingPart.partInfo.title;
                     }
+                    else
+                    {
+                        occludingPart = "Environment";
```
Phantomical: How difficult would it be to make this into `Occluded by <bodyName>` instead when the ray hits a celestial body?

Phantomical (approved):
This looks great! I have one question but otherwise am happy to merge this.

src/Kopernicus/Components/KopernicusSolarPanel.cs
Phantomical: No need to do that, I think. This is good enough as is. Thanks for the PR!

PR #769, opened 2026-02-03
Fix an off-by-1 error in KopernicusCBAttributeMapSO.CompileRGBA

Phantomical:
This was counting down from `_data.Length`, instead of `_data.Length - 1`, so it would perform an OOB read and OOB write with every call.

I don't think this is actually used anywhere but better to fix it anyways.


R-T-B:
Agreed.

PR #777, opened 2026-02-11
Expose Kopernicus version in KSPAssemblyVersion

Phantomical:
Now instead of the KSPAssemblyVersion always being 1.0.0 it is instead 1.0.$(KopernicusVersion). This should break anyone, but it should allow dependencies that want to constrain the kopernicus versions they work with to work correctly.

I figure I might want this for something in the near future, so better to start now.


R-T-B:
As long as you never update the major version this should be fine (there is an ancient, special release for the NASA DART project that uses major version 2, you can read about the collab on the forums somewhere, they did a piss poor job at marketing it lol).

But yeah this looks good.

PR #778, opened 2026-02-14
Optimize away an allocation in PQSTriplanarZoomRotationTextureArray

Phantomical:
Calling `.name` always allocates. This avoids that in some cases by checking it against a known shader instance id, or otherwise checking against the SHADER_NAME constant.

This isn't really a huge difference, but it was showing up on profiles when I am profiling BurstPQS and it is an easy fix


R-T-B:
Isn't PQSTriplanarZoomRotationTextureArray the stock Atlas shader?  I don't think many Kopernicus people use it if so, but still looks good.  At least stock users can kinda benefit (why you'd have Kopernicus in stock, I do not know).

PR #783, opened 2026-02-19
Use memory-mapped textures for MapSODemand

Phantomical:
Right now, when `MapSODemand` is loaded it:
- Loads the entire texture off disk, including uploading it to the GPU
- Copies and/or compiles the texture into a separate large array
- Generally reads on the order of 1 to 64 pixels from that texture at any given time

This is rather inefficient: it uses a ton of memory (especially for compressed textures), requires sending a bunch of data to the GPU that won't be used, and is all around rather slow. Loading large textures is an expensive operation.

To fix this I have been working on a new `CPUTexture2D` in KSPTL. How it works is:
- DDS textures get memory mapped. KSPTL includes support for _all_ the texture formats used by the KSP modding ecosystem.
- Everything else gets loaded as a regular texture, then we call `GetRawTextureData` and use the same texture support as is used for memory mapped textures.

The end result is that the OS can automatically manage which texture pages are resident _and_ that textures loaded this way do not count against the memory usage of the process. Textures in asset bundles are still loaded as textures, but the CPU-side data is just used as-is through `GetRawTextureData` instead of being copied again.

I haven't really done the work to analyze how much RAM this saves, but anecdotally my Sol testing save with 64k textures seems to use about 10Gb less ram so I think the savings should be noticeable. I'm also really happy about with the place this leaves us in. With this PR, a dds texture on disk that is used by both parallax and kopernicus has two copies: one is memory mapped, and one is solely in VRAM. That is the ideal end-state in my mind.

This will stay as a draft for now since it needs the next release of KSPTextureLoader and that isn't released yet.


R-T-B:
Just FYI, I will be building and testing this for bugs.  When you feel it is ready for merge, feel free to ping me.  I will also report any bugs I find.

Phantomical:
Please do! Make sure to use either the KSPTL version included with the PR or the latest prerelease on CKAN. (though there are some bugs in that one). I'm planning on putting out another KSPTL prerelease later tonight that should be more stable and would probably be a good place for you to start testing.

Phantomical:
Uh not sure how I closed this by accident. Will get the commits back and reopen.

Phantomical:
Oh this looks like it got committed over in 8e2061c15e6ac8b11d3d3687283ebd367cbd4b0b

PR #789, opened 2026-02-22
Allow MapSOParsers to handle empty texture names without logging an error

Phantomical:
The pattern of putting
```
texture =
```
in a planet config seems to be a bit of an established practice. Right now this produces error spam from kopernicus that arguably isn't that useful. We might as well just make it an idiom and provide a default texture that can be used.

Since this is meant to be something like an ignored texture the defaults I have picked are:
- `Texture2DParser` - an all magenta texture, which should match unity's missing material shader
- `MapSOParser` - `Greyscale` and `HeightAlpha` are black, everything else is the magenta texture


R-T-B:
This likewise seems sensible heh.  Merging as well and keeping these up to date in my test game.

PR #792, opened 2026-03-02
Add CI workflows for building and attaching release artifacts

Phantomical:
This adds two new CI workflows:
- `build` - which builds a kopernicus release zip for the current commit
- `release` - which builds the release zip (using `build`) and then uploads it as a release artifact.

`build` runs on every commit and pull request, `release` runs whenever a new release is published. It also uploads `Kopernicus.version` as a release artifact. Ideally we should eventually switch to pointing the AVC urls to point at a version artifact in the latest release so we don't need to worry about keeping the one in the repo up to date.

### DIFF
--- a/build/KSP19PLUS/GameData/Kopernicus/Config/00Version.cfg
+++ b/build/KSP19PLUS/GameData/Kopernicus/Config/00Version.cfg
@@ -1,9 +1,9 @@
 KOP_VERSION
 {
-	name = Kopernicus_Version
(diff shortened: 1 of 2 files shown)


Phantomical:
We can probably autogenerate that. I'll add it to the build setup.

R-T-B (approved):
Looks good, had to fix a few old cfgs in the repo but nothing on your end.  Excellent work.

PR #809, opened 2026-03-15
Make another pass at improving `PQSMod_OnDemandHandler`

Phantomical:
This gets `PQSMod_OnDemandHandler` to a point where I don't think there are any more issues with it.

Here's what I have changed:
* Instead of using LateUpdate we now have a coroutine does the check every frame. The coroutine exiting results in OnDemandHandler being considered unloaded. This means that OnDemandHandler instances that are not loaded have no per-frame overhead.
* The coroutine explicitly checks `sphere.active`. `OnSphereInactive` is not reliably emitted during scene switches which was leading to maps not being unloaded appropriately.
* We now track both `_unloadTime` and `_unloadFrame` so that long frames (i.e. scene switches) don't result in maps being unloaded.

I think I can stop making PRs for this after this point.


R-T-B:
Merging this and will test the modern master build as it stands now, thanks.

PR #824, opened 2026-04-13
Upstream part of PQSCoroutineLeakFix into Kopernicus 

Phantomical:
Because we now delay ResetAndWait until after scene switches properly killing the previous ResetAndWaitCoroutine instance is now critical for correctness. When KSPCF is installed it fixes this issue, but when it's not installed you get planets that seem to be stuck at the lowest PQS level with no colliders.


R-T-B:
I have been using this in my test build for a bit without issue so I think its ready for merge.

PR #834, opened 2026-04-30
Adjust the message in the missing texture popup

Phantomical:
It seems like more often then not this popup is showing up due to broken mods, rather than user error. I don't really expect many people to actually read the error message, but this rewords the message to be less alarming.


R-T-B:
Looks good, will merge.  FYI, I plan to try and do a release now, if github holds together.  Pray for us, lol.

PR #844, opened 2026-05-11
Migrate existing shaders and call sites to the new material loader

Phantomical:
This builds on top of #839. It is rather chonky.

It converts every existing per-shader loader from the old loader style to instead inherit from `BaseLoader` and to use the helper functions defined within. All existing configs should continue to work, but now everyone can directly set shader properties by doing `_PropName` and it will automatically pick the right loader by introspecting the shader.


R-T-B:
I'll merge and start testing.

PR #848, opened 2026-05-14
Add support for the `Terrain/Gas Giant` shader

### LINKED ISSUE #831: Allow configuring gas giant shader parameters
KSP's gas giant shaders have a bunch of configurable parameters. Kopernicus exposes none of these for users to configure. We should do so.

Existing documentation for how to do this is here <https://forum.kerbalspaceprogram.com/topic/195178-modders-notes-1100/>

Phantomical:
Adds `ScaledMaterialType.GasGiant` and a GasGiantLoader, wires it into the ScaledVersionLoader dispatch, and attaches the GasGiantMaterialControls component when the gas giant type is in use. Also extends the initial material-type detection to recognize the gas giant shader on existing scaled bodies.

Fixes #831


R-T-B:
Just a note we should document what we can of this on the wiki, but otherwise, this is most excellent work.  Thank you.

PR #850, opened 2026-05-14
Cleanup and simplify loader property definitions

Phantomical:
This is a more general cleanup of a bunch of the loader properties before I start actually making behavioural changes in follow-up PRs. Mostly it replaces explicit fields with the `field` keyword from C# 13.

### DIFF
--- a/src/Kopernicus/Configuration/OceanLoader.cs
+++ b/src/Kopernicus/Configuration/OceanLoader.cs
@@ -169,16 +169,13 @@ namespace Kopernicus.Configuration
             }
         }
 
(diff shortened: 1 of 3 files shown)


R-T-B:
There are some conflicts here, would you like me to sort them, or you?  Happy to merge this.  Let me know when would be a good time to take a "testing snapshot" of your changes for my livegame and testbed.

Phantomical:
I'll sort out the merge conflicts. There's a few more sets of changes after this, then it should be in a good coherent state to test.

PR #855, opened 2026-05-17
Allow material loaders to configure the parent object 

Phantomical:
We have a bunch of cases where a shader needs a custom support game object in order to handle stuff at runtime. We already special-case a couple of different shaders, and if we want to allow custom shaders then we can't just add that all the appropriate PostApply.

This adds a new interface and a few different callbacks on BaseMaterialLoader that get called at during PostApply for each parent
loader. This lets us move anything that is special for a specific shader out of PQSLoader.

Depends on #854


R-T-B:
Making a testing build as of this merge, just FYI.  I'll provide feedback.

PR #897, opened 2026-06-21
Add German Localization

Oprissmian:
Adds German localization for Kopernicus_SolarPanelFixer and Kopernicus_UI_MissingTextures


Phantomical:
Thanks!

Issue #825, opened 2026-04-14
Strange behavior with fast rotators


Clayell:
Video: https://youtu.be/7KzQNt_tLsw?t=27
Logs: https://drive.google.com/drive/folders/1UYimztFGA35f0enF5H9KuIKHA_hpkQ6c?usp=sharing
2 different bugs are present here:

1. (0:27 to 1:25) The vessel's projected orbit around the sun is changing while still in the SOI of the body, despite the fact that it should no longer be changing. This seems to be because the direction of the velocity is somehow linked to the rotation of the planet? The orbital velocity is also increasing, which seems to be because the surface horizontal velocity is staying at 0 which it definitely should not be
2. (starting at 1:26) After exiting the SOI of the body, the LAN of the vessel's orbit is somehow linked to the rotation rate of the body, despite the fact that the body should have no influence on the vessel anymore

The bug goes away after reloading the scene, as seen in the video

To reproduce:

The mod I'm using is a modified version of Scallop's Extra Moon, with the modified config below:
```
@Kopernicus:AFTER[KOPERNICUS] 
{
    Body
    {
        name = Scylla
        Template
        {
            name = Gilly
        }
        Properties
        {
            description = A small orbiting Kerbin beyond the orbit of Minmus, and is not visable from the surface.
            radius = 5472
            geeASL = 0.000000015
            rotationPeriod = 150
			ScienceValues
			{
				landedDataValue = 3
				inSpaceLowDataValue = 2
				inSpaceHighDataValue = 1
				recoveryValue = 5		
			}
        }
        ScaledVersion
        {
            Material
            {	
                texture = Scylla/Scylla.png
				normals = Scylla/Normals.png
            }
        }
        Orbit
        {
            referenceBody = Sun
            color = 1,0,0,1
            inclination = 5
            eccentricity = 0
            semiMajorAxis = 200000000000
  
        }
        PQS
        {
            Mods
            {          		
				VertexColorMap
                {
                    map = Scylla/Scylla.png
                    order = 9999999
                    enabled = true
                }
				VertexHeightMap
                {
                    map = Scylla/Scylla.png
                    deformity = 14000
                    order = 9999999
                    enabled = true
                }
            }
        }
    }
}
```

CKAN mod list: (Extra Moon + Kopernicus and dependencies)
<img width="1418" height="771" alt="Image" src="https://github.com/user-attachments/assets/5e9a3fe6-7d8b-4456-b699-1b653672b74b" />

[...]

PR #904, opened 2026-07-19
Clear outgoing body's inverseRotation flag

ballisticfox:
When a craft leaves the SOI of a moon but is still in the inverse rotation threshold of the object, `OrbitPhysicsManager.setDominantBody()` reassigns `dominantBody` but doesn't clear the outgoing body's `inverseRotation` flag.

`Planetarium.CBUpdate()` rebuilds the single global `Planetarium.Zup` frame from any body with inverseRotation == true, and that frame is what every on-rails body position is expressed through (`getPositionFromTrueAnomaly` -> `Planetarium.Zup.WorldToLocal`).

This closes half of issue #825 (the craft being dragged around the sun).


R-T-B:
Good catch, lgtm.

PR #907, opened 2026-08-14
Extend asteroid size configuration options

siimav:
1. Asteroid groups can now set the physical radius of each size class directly, in meters, via a `ClassRadius` node.
   Classes left out of the node keep the stock scaling.
2. Asteroid groups can now set `minClass`/`maxClass` to restrict which size classes they generate.
   Defaults (A and I) preserve existing behaviour; note that stock's own spawner only goes up to E.
3. Fixed a `Size` curve that evaluates to exactly 1 producing an out-of-range size class, which
   resolved to a nonexistent `Procedural/PA_9` prefab and left the object with no model or mass.

```
    Asteroid
    {
        name = Stock

        //  Optional. Restricts which size classes this group generates: the Size curve below maps
        //  its [0,1] output onto minClass..maxClass. Defaults are A and I. Note that Squad's own
        //  spawner only goes up to E, so set maxClass = E to match the stock class distribution.
        minClass = A
        maxClass = E

        //  Optional. Physical radius per size class, in meters. Without it, an object's radius is
        //  the radius baked into the PA_<class> prefab (~3.2m at class A up to ~193m at class I)
        //  scaled by minRadiusMultiplier/maxRadiusMultiplier from the PotatoRoid part's
        //  ModuleAsteroid. Those two fields are shared by every class, so once their spread is
        //  wider than the ~1.67x gap between adjacent classes the classes overlap in size and the
        //  class letter stops telling the player anything. Setting a range here pins that class to
        //  it instead. Any class left out keeps the stock scaling.
        ClassRadius
        {
            A
            {
                minValue = 2
                maxValue = 6
            }
            B
            {
                minValue = 6
                maxValue = 15
            }
        }
    }
```


src/Kopernicus/Configuration/DiscoverableObjects/Asteroid.cs
```diff
@@ -90,6 +90,24 @@ private List<NumericCollectionParser<Single>> SizeSetter
 
         public FloatCurve Size { get; set; }
```
Phantomical, line 93: I think the docs on the size curve would be better off living on `Size` itself.

```suggestion
        // A float curve that governs how likely each asteroid size and class is to be generated.
        // It should range from 0 to 1, with 0 mapping to MinClass and 1 to MaxClass.
        public FloatCurve Size { get; set; }
```

src/Kopernicus/Configuration/DiscoverableObjects/Asteroid.cs
```diff
@@ -90,6 +90,24 @@ private List<NumericCollectionParser<Single>> SizeSetter
...
 
+        // The smallest size class this group generates. The Size curve maps its [0,1] output onto
+        // the MinClass..MaxClass range, so narrowing the range here concentrates the distribution
+        // instead of spreading it over every class the game defines.
```
Phantomical:
```suggestion
        // The smallest size class that can be generated by this group. Defaults to A.
```

src/Kopernicus/Configuration/DiscoverableObjects/Asteroid.cs
```diff
@@ -90,6 +90,24 @@ private List<NumericCollectionParser<Single>> SizeSetter
...
+        [ParserTarget("minClass")]
+        [SuppressMessage("ReSharper", "FieldCanBeMadeReadOnly.Global")]
+        public EnumParser<UntrackedObjectClass> MinClass = UntrackedObjectClass.A;
+
+        // The largest size class this group generates. Defaults to I so the class spread matches
+        // what Kopernicus has always produced; stock's own spawner only goes up to E by default.
```
Phantomical:
```suggestion
        // The largest size class that can be generated by this group. Defaults to I. Note that stock
        // only spawns asteroids up to E by default.
```

src/Kopernicus/Configuration/DiscoverableObjects/Asteroid.cs
```diff
@@ -90,6 +90,24 @@ private List<NumericCollectionParser<Single>> SizeSetter
...
+        [ParserTarget("maxClass")]
+        [SuppressMessage("ReSharper", "FieldCanBeMadeReadOnly.Global")]
+        public EnumParser<UntrackedObjectClass> MaxClass = UntrackedObjectClass.I;
+
+        // Physical radius per size class, in meters. Optional - any class left out keeps the stock
+        // ModuleAsteroid scaling. See ClassRadiusLoader for why this exists.
```
Phantomical:
```suggestion
        // Asteroid physical radius configuration. See ClassRadiusLoader for details. Defaults
        // to the stock asteroid radius scaling if not specified.
```

src/Kopernicus/Configuration/DiscoverableObjects/ClassRadiusLoader.cs
```diff
@@ -0,0 +1,107 @@
...
+    // minRadiusMultiplier/maxRadiusMultiplier fields on ModuleAsteroid. Those multipliers live on
+    // the part, so every class shares them, and the radii baked into the prefabs (~3.2m for class
+    // A up to ~193m for class I) are the only thing separating one class from another. Once the
+    // multiplier spread is wider than the ~1.67x gap between adjacent classes, the classes overlap
+    // and the class letter stops saying anything useful about the object's size.
+    //
+    // Giving a class a range here lets a config state the radius it actually wants. Kopernicus
+    // divides out the prefab radius and hands stock the multipliers that produce that range, so
+    // the roll still happens in stock code off the persisted seed. Classes left unset are not
+    // touched at all and keep the stock scaling.
```
Phantomical:
```suggestion
    // Radius configuration for asteroid size classes.
    //
    // Normally, the asteroid size multipliers is controlled by the stock prefab for each individual
    // size class. This config loader allows you to specify what radii you want for each asteroid
    // class, then Kopernicus take care of making that apply to any asteroids spawned from that
    // class.
    //
    // Actually picking a number is still done by stock code and unset classes keep the same scaling as stock.
```

src/Kopernicus/RuntimeUtility/DiscoverableObjects.cs
```diff
@@ -311,6 +311,26 @@ public void SpawnAsteroid(Asteroid asteroid, UInt32 seed)
             Debug.Log("[Kopernicus] New object found near " + body.name + ": " + protoVessel.vesselName + "!");
         }
 
+        // Picks the size class for a new object by mapping the group's Size curve onto its
+        // [minClass, maxClass] range. The result is clamped rather than cast straight from the
+        // curve: a curve that evaluates to exactly 1 used to produce an out-of-range enum value,
+        // which stock then turned into a prefab URL ("Procedural/PA_9") that doesn't resolve.
```
Phantomical:
```suggestion
        // Picks the size class for a new object by mapping the group's Size curve onto its
        // [minClass, maxClass] range.
```

src/Kopernicus/RuntimeUtility/DiscoverableObjects.cs
```diff
@@ -332,6 +352,66 @@ private IEnumerator<WaitForSecondsRealtime> AsteroidDaemon(Asteroid asteroidGrou
 
         public static string LaunchedFromName(Asteroid asteroid) => $"AST-{asteroid.Name}";
 
+        // Maps the launch site name stamped onto every object we spawn back to the group that
+        // spawned it. Vessel.launchedFrom is what stock persists for us - a non persistent
+        // PartModule field written into the spawned vessel node would survive the first load and
+        // then be dropped the next time the game saved, so this is the only channel that holds.
+        // Built lazily off the config rather than in AsteroidSetup so that a vessel loading early
+        // in a scene can't beat this scenario's Start() to it.
+        private static Dictionary<String, Asteroid> _groupsByLaunchSite;
```
Phantomical: Is this actually grouping by launch site? As far as I can tell it is more or less grouping by asteroid name? I don't get what the key in this map is supposed to be, because the comment is confusing to me.

Beyond that, I think this should maybe be built in a `PostApply` at config parse time? It looks like it is only ever built once.

src/Kopernicus/RuntimeUtility/DiscoverableObjects.cs
```diff
@@ -332,6 +352,66 @@ private IEnumerator<WaitForSecondsRealtime> AsteroidDaemon(Asteroid asteroidGrou
...
+                if (range.MinValue.Value <= 0f || range.MaxValue.Value <= 0f)
+                {
+                    Debug.LogWarning("[Kopernicus] Asteroid group " + asteroid.Name + " has a non-positive ClassRadius for class " +
+                                     sizeClass + "; that class falls back to the stock scaling.");
+                }
+                else if (range.MaxValue.Value < range.MinValue.Value)
+                {
+                    Debug.LogWarning("[Kopernicus] Asteroid group " + asteroid.Name +
+                                     " has ClassRadius minValue > maxValue for class " + sizeClass + ".");
+                }
```
Phantomical, line 410: I think these should probably be using the kopernicus log error functions so that they get printed to both `X.Body.log` and `KSP.log`.

src/Kopernicus/Patches/ModuleAsteroid_OnStart.cs
```diff
@@ -0,0 +1,79 @@
...
+namespace Kopernicus.Patches;
+
+// Applies the per-class radius ranges from an asteroid group's ClassRadius node.
+// Stock computes the object's radius in ModuleAsteroid.OnStart as
+//   radius = paPrefab.radius * Random.Range(minRadiusMultiplier, maxRadiusMultiplier)
+// where paPrefab.radius is baked into the PA_<class> Unity prefab and the two multipliers are
+// PartModule fields shared by every class. Writing the multipliers here, before the original
+// method runs, lets a config ask for a radius in meters per class without touching stock's own
+// logic: the roll still happens in stock code, off Random.InitState(seed) with the persisted
+// seed, so an object's size stays a pure function of its seed and the config.
```
Phantomical:
```suggestion
// Apply configured asteroid class radius ranges to the asteroid itself.
//
// Stock computes the radius in ModuleAsteroid.OnStart as
//    radius = paPrefab.radius * Random.Range(minRadiusMultiplier, maxRadiusMultiplier)
// We override the min/max radius multipliers here in the prefab to be
//    minRadiusMultiplier = paPrefab.radius / minRadius;
//    maxRadiusMultiplier = paPrefab.radius / maxRadius;
// which makes stock code pick from the radius range we want it to.
```

Phantomical (commented):
I took a shot at rewriting the doc comments from what is (I think?) claudish to something that works a little better.

I have some questions about how exactly the launchedFrom bit is meant to work and whether kopernicus is already setting `launchedFrom` in some way - it is possible I just need to read the existing code more here, though.

Other than that one small nit on where warning messages are printed in case of an invalid config.

src/Kopernicus/RuntimeUtility/DiscoverableObjects.cs
siimav: We do write $"AST-{asteroid.Name}" onto the launchedFrom field. The comment is mostly about how we shove launchedFrom value into the protovessel confignode when spawning the asteroid and KSP will take over from there. I guess the note about non-persistent partmodule field can be confusing when you don't have the full context.

src/Kopernicus/RuntimeUtility/DiscoverableObjects.cs
siimav: Hmm, but the rest of the code here appears to use Unity logger. Loader.cs does have some kind of per-body logging so you'd like to get something similar smuggled in here?

src/Kopernicus/RuntimeUtility/DiscoverableObjects.cs
siimav: Oh I see the confusion. `Asteroid.Name` isn't actually the name of the asteroid. `Asteroid` is in fact the Kopernicus config for asteroid group. Yes, it really is a bad name.

src/Kopernicus/RuntimeUtility/DiscoverableObjects.cs
Phantomical: If the rest of the file is using that then let's stay consistent for now. I'll create a follow-up issue to move the whole file over.

src/Kopernicus/RuntimeUtility/DiscoverableObjects.cs
```diff
@@ -332,6 +350,59 @@ private IEnumerator<WaitForSecondsRealtime> AsteroidDaemon(Asteroid asteroidGrou
```
Phantomical, line 350: I think as long as we explain the group -> `LaunchedFromName(group)` -> `Vessel.launchedFrom` bit I'll be happy with this.
```suggestion

        /// <summary>
        /// Compute the name stored <see cref="Vessel.launchedFrom" /> for an asteroid group.
        /// </summary>
        /// <remarks>
        /// We don't really have a good way to track what group an asteroid belongs to within
        /// Kopernicus, so instead we store this in <see cref="Vessel.launchedFrom" /> for asteroids
        /// so that their group can be recovered later on.
        /// </remarks>
```

Phantomical (approved):
LGTM with one more comment request. Otherwise I believe @R-T-B wanted to review.

src/Kopernicus/RuntimeUtility/DiscoverableObjects.cs
R-T-B: Yes logging is inconsistently used in Kopernicus.  I've been meaning to move everything to the Kopernicus logger for some time, but yeah, it's not a big concern at the moment for this merge.  Just stay consistent.

R-T-B:
LGTM as well.  I'll be happy to merge this when you are done with whatever you are doing right now (draft status).

siimav:
I marked it as draft because the only testing it got was loading into the save I had where a vessel was about to claw at a 🥔. Worked fine in that instance but not sure what your validation requirements are.

R-T-B:
Ah, I see.  I will test it in my personal testing ground savegame then.

ballisticfox:
I'll do some investigation tonight with Sol if I happen to get the time

Phantomical:
Marking this as being ready for review with the understanding that some testing still needs to be done.

siimav:
Squeezed in one more quick test. Everything worked as expected.

R-T-B:
Working ok here as well.

R-T-B:
I forgot I was even testing with this lol, but anyways it worked in my brief stint with playing with the options and has caused no issues in my live savegame for nearly a month, so I think a merge now is safe.

PR #910, opened 2026-08-17
Set both deactivateAltitude and highQualityShaderDeactivateAltitude

ballisticfox:
Apparently there's a second hidden PQS deactivation altitude variable called highQualityShaderDeactivateAltitude which is seemingly only used by Bop for whatever reason. This now sets both highQualityShaderDeactivateAltitude and deactivateAltitude to be the same value.


R-T-B:
lgtm, as discussed I think this is the right approach.  Merging.

PR #913, opened 2026-08-23
Custom Ring Shader Loader

ballisticfox:
Revamps the ring shader system to use something more akin to what ScaledSpace and PQS have set up.

This probably needs more testing, but I know it seems to load rings using NewShader without any of the detailTiling features, and rings with custom shaders without problems.

### DIFF
--- a/src/Kopernicus/Components/Ring.cs
+++ b/src/Kopernicus/Components/Ring.cs
@@ -26,8 +26,10 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using Kopernicus.Components.ModularComponentSystem;
 using Kopernicus.Components.Serialization;
+using Kopernicus.OnDemand;
 using UnityEngine;
 
 namespace Kopernicus.Components
@@ -59,8 +61,6 @@ namespace Kopernicus.Components
 
         public Single planetRadius;
         public Quaternion rotation;
-        public Texture2D texture;
-        public Color color;
         public Boolean lockRotation;
 
         /// <summary>
@@ -76,13 +76,18 @@ namespace Kopernicus.Components
         /// </summary>
         public Single longitudeOfAscendingNode;
 
-        public Boolean unlit;
-        public Boolean useNewShader;
+        /// <summary>
+        /// The ring's material, built by RingLoader from the config. Everything a config can author
+        /// lives on it; this component writes only what it derives from the mesh and from where the
+        /// star is, which is why those are the sole properties it still names below.
+        /// </summary>
+        public Material material;
 
-        public Single albedoStrength = 1.0F;
-        public Single scatteringStrength = 1.92466F;
-        public Single anisotropy = 0.95F;
-        public Texture2D backlitTexture;
+        /// <summary>
+        /// Textures the material deferred to on-demand loading, as shader property name to texture
+        /// path. Carried here rather than wired at parse time — see BuildRing.
+        /// </summary>
+        public Dictionary<String, String> materialOnDemandTextures;
 
         public Single radiusMultiplier = 1.0F;
 
@@ -94,17 +99,6 @@ namespace Kopernicus.Components
         /// from (0,0) to (1,1).
         /// </summary>
         public Int32 tiles;
-
-        /// <summary>
-        /// For new shader, makes planet shadow softer (values larger than one) or less soft (smaller than one)
-        /// softness still depends on distance from sun, distance from planet and radius of sun and planet
-        /// </summary>
-        public Single penumbraMultiplier = 10f;
-
-        /// <summary>
-        /// This texture's opaque pixels cast shadows on our inner surface
-        /// </summary>
-        public Texture2D innerShadeTexture;
 
         /// <summary>
         /// The inner shade texture repeats this many times over the inner surface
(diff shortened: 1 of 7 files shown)


Phantomical:
Once this is ready move it out of draft and I'll review. One thing I would definitely like to see here, though, is on-demand enabled for rings. There are a few mods with enormously large ring textures (600MB+ vram) so ideally we can get that working properly.

You don't have to do that, of course. If not, I'll see if I can do a follow up to get it working and enabled by default.

ballisticfox:
I'll take a look at it. Also if you don't mind, could you skim over the code and make sure I didn't do anything egregious?

src/Kopernicus/Components/Ring.cs
```diff
@@ -404,19 +275,10 @@ public void BuildRing()
         }
 
         /// <summary>
-        /// The shaders used by the ring mesh
+        /// Fallback shader for a ring built in code. Which shader a parsed ring gets is decided by
+        /// its config, in RingLoader.
```
Phantomical, line 279:
```suggestion
        /// Default ring shader.
```

src/Kopernicus/Configuration/MaterialLoader/RingsLoader.cs
```diff
@@ -0,0 +1,157 @@
...
+        // The ring itself, sampled across its width. Alpha is opacity.
+        [ParserTarget("texture")]
+        public MaterialTextureParser MainTexture
+        {
+            get => GetTextureName("_MainTex");
+            set => SetTexture("_MainTex", value);
+        }
+
+        // Sampled instead of `texture` where the ring is lit from behind. Defaults to `texture`, so a
+        // ring that sets neither still renders and one that sets only `texture` keeps its old look.
```
Phantomical, line 60:
```suggestion
        // Texture to use when the ring is lit from behind. If not specified it defaults to whatever
        // was specified for `texture`.
```

src/Kopernicus/Configuration/RingLoader.cs
```diff
@@ -299,154 +294,25 @@ public NumericParser<Single> InnerShadeRotationPeriod
...
-            {
-                get { return new DetailPassLoader(Value.fine); }
-                set { Value.fine = value.Value; }
-            }
-        }
+        // Material of our ring
+        [ParserTarget("Material", AllowMerge = true, GetChild = false)]
+        [KittopiaUntouchable]
+        [KittopiaDescription("Shader and material properties for the ring. Omit to keep the built-in "
+                             + "shader chosen by useNewShader/unlit.")]
```
Phantomical, line 314:
```suggestion
        [KittopiaDescription("Shader and material properties for the ring.")]
```

src/Kopernicus/Configuration/MaterialLoader/RingsLoader.cs
```diff
@@ -0,0 +1,157 @@
...
+        public const string SHADER_NAME = "Kopernicus/Rings";
+
+        public static bool UsesSameShader(Material m) => m != null && m.shader != null && m.shader.name == SHADER_NAME;
+
+        [PreApply]
+        [ParserTarget("shader")]
+        public override ShaderParser ShaderParser { get; set; }
+
+        [ParserTarget("onDemand")]
+        public override NumericParser<bool> OnDemand { get; set; } = false;
```
Phantomical, line 49: I think we should leave this at the non-override default here, that way it gets controlled by the kopernicus defaults.

src/Kopernicus/Components/Ring.cs
```diff
@@ -334,62 +222,45 @@ public void BuildRing()
...
-                    ringMr.sharedMaterial.SetTexture(BacklitTexture, backlitTexture);
-                }
-                if (backlitTexture == null)
-                {
-                    ringMr.sharedMaterial.SetTexture(BacklitTexture, texture);
-                }
+                onDemandLoader.Entries = materialOnDemandTextures
+                    .Select(kv => new OnDemandTextureEntry(kv.Key, kv.Value))
+                    .ToList();
+            }
```
Phantomical, line 245: This seems to be reimplementing `MaterialLoader.OnParentApply`. Just call that instead, `ScaledMaterialLoader` overrides it to do what you want here.

src/Kopernicus/Configuration/MaterialLoader/RingsLoader.cs
```diff
@@ -0,0 +1,157 @@
...
+            if (material != null)
+                ShaderParser = material.shader;
+        }
+        public void ApplyDeferred()
+        {
+            if (Value == null)
+                return;
+
+            if (Value.HasProperty("_BacklitTexture") && Value.GetTexture("_BacklitTexture") == null)
+                Value.SetTexture("_BacklitTexture", Value.GetTexture("_MainTex"));
```
Phantomical, line 152: As-is this doesn't work if on-demand is enabled for this material.
```suggestion
          if (Value.HasProperty("_BacklitTexture") && Value.GetTextureName("_BacklitTexture") == null)
                Value.SetTexture("_BacklitTexture", Value.GetTextureName("_MainTex"));
```

Phantomical (commented):
I've had these pending for a bit so I figure I should just post them. I have more comments that I need to make, but there are some issues that need to be addressed from these as well.

